PR #28, opened 2021-03-24
Update/time picker

j-mnizhek:
- Remove icons for AM/PM because they confused some of our users
- Remove fixed-width for AM/PM button
- Update guide and illustrations


j-mnizhek:
Do we need to update snapshots for this component?
My locale tests are always failing for all components for some reasons 🤔

PR #31, opened 2021-04-12
[D3] Add Donut chart

lsroman:
Обновился Popper и popper.js
Переименовался YXPlot -> Chart


semcore/d3-chart/src/Donut.js
```diff
@@ -0,0 +1,127 @@
...
+
+import style from './style/donut.shadow.css';
+
+const DEFAULT_INSTANCE = Symbol('DEFAULT_INSTANCE');
+
+class DonutRoot extends Component {
+  static displayName = 'Donut';
+  static style = style;
+
+  static defaultProps = ({ innerRadius = 0, half = false, $rootProps: { size } }) => {
```
ekrasnopolskaia, line 17: maybe semi-circle instead of half

semcore/d3-chart/src/Donut.js
```diff
@@ -0,0 +1,127 @@
...
+
+function Label({ Element: SLabel, styles, Children }) {
+  return styled(styles)(
+    <SLabel render="text" x="0" y="0">
+      <Children />
+    </SLabel>,
+  );
+}
+
+const Donut = createElement(DonutRoot, { Pie, Label, NotData: Pie });
```
ekrasnopolskaia, line 125: maybe not NotData, but NullData? It's more consistent 
or just Null, like in other charts

PR #102, opened 2021-10-12
Fix react key-related warning

dpodolkhov:
Avoid react warning: Each child in a list should have a unique "key" prop. Check the render method of `Bar`...


dpodolkhov:
SBar был обернут во фрагмент, на котором не было key.

PR #164, opened 2022-03-03
fixed plot margin cut

phytonmk:
Before:
index-test-js-bar-chart-should-not-cut-content-on-right-with-left-margin-1-snap
After:
index-test-js-bar-chart-should-not-cut-content-on-right-with-left-margin-1-snap-1

I have not enough context and existing tests. @lsroman, review runtime changes please

Internal issue source: https://semrush.slack.com/archives/C293ALLQ2/p1645699001536649


lsroman:
Looks good 👍

lsroman (approved):

PR #165, opened 2022-03-03
Disabled animation if reduce motion is preferred.

phytonmk:
PR cause it will trigger publishing a half of out packages


phytonmk:
fuck, also need to disable transitions

phytonmk:
All done. It will be 55 packages to be published. @lsroman, do u confirm another one whole monorepo publish?

PR #241, opened 2022-05-19
[chore] removed legacy @semcore/babel-plugin-shadow

phytonmk:
@lsroman, is it safe action?


phytonmk:
For history: totally not safe for now

PR #295, opened 2022-07-03
Accessible charts

phytonmk:
What changed:
Added a11y module for `@semcore/d3-chart` package.

Code is ready for review.
Pull request contains data and charts example that should be removed before merge.

Risks:
- Charts summary may be not generated for some edge cases
- a11y async import may not work in some bundling systems

What next:
- To add user documentation
- To test on some semrush service
- Add more translations (only English is available for now)

Architecture is clear as the code but also will be described in the user documentation


semcore/d3-chart/src/a11y/locale.ts
```diff
@@ -0,0 +1,27 @@
...
+    const [localeBase] = locale.split('-');
+    if (translations[localeBase]) {
+      locale = localeBase;
+    }
+  }
+  locale = translationNames[locale] ? translationNames[locale] : locale;
+  if (!translations[locale]) {
+    const availableLocales = Object.keys(translations).join(', ');
+    // eslint-disable-next-line no-console
+    console.error(
```
lsroman, line 21: Let's do it through our logger because there is a check for `process.env.NODE_ENV !== 'production'`

semcore/d3-chart/src/a11y/PlotA11yModule.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const hadnleHiddenElementsFocus = React.useCallback(() => {
+    setWasFocused(true);
+    setNavWithKeyboard(true);
+  }, []);
+
+  const [loading, setLoading] = React.useState(false);
+  const [error, setError] = React.useState<Error | null>(null);
+
+  const locale = React.useMemo(() => normalizeLocale(props.locale, translations), [props.locale]);
```
lsroman, line 38: Let's also take language out of context. Many applications are already wrapped in this.
import { I18nConsumer } from '@semcore/utils/lib/enhances/I18nConsumer';

semcore/d3-chart/src/a11y/PlotA11yView.tsx
```diff
@@ -0,0 +1,182 @@
...
+      summaryPlaceholder: intl.formatMessage({ id: 'view-summary-placeholder' }),
+      table: intl.formatMessage({ id: 'view-table' }),
+    }),
+    [intl, plotLabel],
+  );
+
+  const dataTableRef = React.useRef<HTMLTableElement>(null);
+
+  return sstyled(styles)(
+    <SPlotA11yView render="div" tabIndex={0} aria-label={texts.label}>
```
lsroman, line 143: I understand that this is not what we need, but let's make a render from the Box component because I see a lot of errors in the console.
Screenshot 2022-08-19 at 14 49 21

lsroman (commented):
This is great work. I checked the asynchrony and all popular builder tools deliver the chunk perfectly 🔥

PR #306, opened 2022-07-12
[input-number] fixed rounding float numbers

freeyoungstrong:


semcore/input-number/src/InputNumber.jsx
```diff
@@ -77,6 +77,7 @@ class Value extends Component {
 
   handleValidation = (e) => {
     const { value, min, max, step } = this.asProps;
+    const roundCoefficient = (step < 1 && step.toString().split('.')[1].length) || 1;
```
phytonmk, line 80: neat: `x && y || z` seems to be very similar to `x ? y : z`

PR #314, opened 2022-07-19
[builder] added mjs output to to cjs files

phytonmk:
Fixed building via rollup (especially with vite)


phytonmk:
@lsroman, do you approve changes?

PR #316, opened 2022-07-20
[icon] fixed typo for AppsBlock icon name

j-mnizhek:


phytonmk:
Proposed changes will make both `@semcore/ui` and `@semcore/icon` new version backward incompatible. I will add deprecated fallbacks for the old name.

Please consider any feature renaming very undesirable and double check icon names before adding them.

PR #341, opened 2022-08-12
[button] added empty button aria-label check

freeyoungstrong:


website/docs/components/button/components/playground.jsx
```diff
@@ -111,6 +111,7 @@ const Preview = (preview) => {
       loading={loading}
       disabled={disabled || loading}
       active={active}
+      ariaLabel="Text"
```
phytonmk, line 114: We have not to popularize bad practice to put meaningless labels in our examples. Please  add prop to editable by user in playground

semcore/button/src/index.d.ts
```diff
@@ -29,6 +29,10 @@ export interface IButtonProps extends IBoxProps, INeighborItemProps, IKeyboardFo
   addonLeft?: React.ElementType;
   /** Tag for the right Addon */
   addonRight?: React.ElementType;
+  /** aria-label attribute for a11y
```
phytonmk, line 32: Why have you put this line outside of `@description`?

website/docs/components/button/examples/one-addon.jsx
```diff
@@ -4,7 +4,7 @@ import CheckM from '@semcore/icon/Check/m';
 
 export default function () {
   return (
-    <Button>
+    <Button ariaLabel="check icon">
```
phytonmk, line 7: That text is very technical and looks like we are suggesting developers to use techincal and meaningless labels

website/docs/components/counter/examples/counter-button.jsx
```diff
@@ -5,7 +5,7 @@ import SettingsXS from '@semcore/icon/Settings/m';
 
 export default () => (
   <>
-    <Button mr={4}>
+    <Button mr={4} ariaLabel="Manage columns">
```
phytonmk, line 8: How columns management is related to `23`? (double times)

semcore/button/src/index.d.ts
```diff
@@ -29,6 +29,10 @@ export interface IButtonProps extends IBoxProps, INeighborItemProps, IKeyboardFo
...
+   * @description This prop will be required in the next major release
+   * */
+  ariaLabel?: string;
```
phytonmk, line 35: As far as `ariaLabel` looks very similar to `aria-label` but still differs, we need either to name it in the isomorphic way `label` or just `aria-label` to redefine and enforce html attribute usage

phytonmk (changes requested):
1. `ariaLabel` name should be changed (more details in comment of `.d.ts` file change).
2. Almost all examples labels should be changed. In our examples we should define the way of new property – describe what the button is about for end user. E.g. "Confirm" is both short and informative description for button containing check icon. Current examples are very abstract and may either confuse developers or make them think about `label` as a technical property.

semcore/button/src/index.d.ts
freeyoungstrong: to be honest idk. I just wanted to show users that this prop will be required in the next release and put in a separate field

lsroman:
> 1. `ariaLabel` name should be changed (more details in comment of `.d.ts` file change).
> 2. Almost all examples labels should be changed. In our examples we should define the way of new property – describe what the button is about for end user. E.g. "Confirm" is both short and informative description for button containing check icon. Current examples are very abstract and may either confuse developers or make them think about `label` as a technical property.

1. Let's not give a description of the standard html attributes. It's a road to nowhere
2. It seems redundant to me to give such examples. 
`<Button type="submit" ariaLabel="Submit">Submit</Button>`
I don't have to describe to do `ariaLabel="Submit` this should work automatically.

lsroman:
Here is a tool for getting text
https://github.com/semrush/intergalactic/pull/344/files#diff-47de7dd65e83cf2998f129706732fc94a6f7491dd5f4de82ec0cb196412fb947

lsroman:
Good job 🔥 thank you for correcting many examples on the site!

PR #344, opened 2022-08-12
[badge,utils] Add support a11y

lsroman:
…onvert react component to text.

Signed-off-by: r.lysov <r.lysov@semrush.com>

Added support a11y for Badge component

## Description

- Add aria-hidden for component
- Added util function `reactToText` to convert react component to text
- Update documentation

## Motivation and Context

The component becomes accessible 💪

## How has this been tested?

Turn on the screen reader and listen to how it reads label text.

## Types of changes

## Checklist:

### DIFF
--- a/semcore/badge/src/Badge.jsx
+++ b/semcore/badge/src/Badge.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import createComponent, { Component, Root, sstyled } from '@semcore/core';
 import { Box } from '@semcore/flex-box';
 import resolveColor from '@semcore/utils/lib/color';
+import reactToText from '@semcore/utils/lib/reactToText';
 
 import style from './style/badge.shadow.css';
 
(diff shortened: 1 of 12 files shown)


semcore/badge/src/Badge.jsx
```diff
@@ -15,10 +16,16 @@ class RootBadge extends Component {
...
     const SBadge = Root;
-    const { styles, color, bg } = this.asProps;
+    const { styles, color, bg, children } = this.asProps;
 
     return sstyled(styles)(
-      <SBadge render={Box} tag="span" use:color={resolveColor(color)} use:bg={resolveColor(bg)} />,
+      <SBadge
+        render={Box}
+        tag="span"
+        aria-label={`badge: ${reactToText(children)}`}
```
phytonmk, line 25: What `badge` means for components end user?

lsroman: it seems to me that for the voice acting it will be better to sound "position tracking icon new" than "position tracking new". But maybe I'm wrong. We need the opinion of UX specialist

PR #349, opened 2022-08-16
[feedback-form, icon] added a11y

freeyoungstrong:
added 11y to FeedbackForm example and Notice component


semcore/icon/src/Icon.jsx
```diff
@@ -35,11 +35,23 @@ function Icon(props, ref) {
     keyboardFocused: keyboardFocused,
   });
 
+  function onKeyDown(event) {
+    if (props.onKeyDown) {
+      props.onKeyDown();
```
phytonmk, line 40: Do not recall default handlers directly, they are already implicitly binded by babel plugins (in short: remove this line)

phytonmk (approved):

PR #356, opened 2022-08-19
[notice, notice-bubble, notice-global] added attributes for better accessibility

freeyoungstrong:
added aria-live attribute
added aria-hidden attribute in docs (examples)


semcore/notice/src/Notice.jsx
```diff
@@ -50,6 +50,7 @@ class RootNotice extends Component {
         use:theme={useTheme}
         backgroundColor={color}
         role="alert"
+        aria-live="assertive"
```
lsroman, line 53: it seems to me that `aria-live="assertive"` can only be used for the error type.
And for the rest it should be `aria-live="polite"`.
If so, then the alert role is not needed.

website/docs/components/notice/components/playground.jsx
```diff
@@ -68,7 +68,7 @@ const Preview = (preview) => {
   return (
     <Notice theme={theme} hidden={hidden}>
       {label && (
-        <Notice.Label>
+        <Notice.Label aria-hidden={true}>
```
lsroman, line 71: Seems like an extra step. The icon will be aria-hidden anyway

lsroman (changes requested):
Let's discuss the changes in voice. It will be faster.)

PR #358, opened 2022-08-22
[tag, input-tags] added accessibility

ekrasnopolskaia:
## Types of changes

## Checklist:

### DIFF
--- a/website/docs/components/input-tags/examples/basic.jsx
+++ b/website/docs/components/input-tags/examples/basic.jsx
@@ -1,6 +1,5 @@
 import React, { useState } from 'react';
 import InputTags from '@semcore/input-tags';
-import Tag from '@semcore/tag';
 import Tooltip from '@semcore/tooltip';
 
 const Demo = () => {
@@ -20,6 +19,13 @@ const Demo = () => {
 
   const handleCloseTag = (e) => {
     e.preventDefault();
+  };
+
+  const handleKey = (e) => {
+    if (e.code === 'Enter' || e.code === 'Space') {
+      handleEditTag(e);
+      handleCloseTag(e);
+    }
   };
 
   const handleEditTag = (e) => {
@@ -50,6 +56,7 @@ const Demo = () => {
             editable
             data-id={idx}
             onClick={handleEditTag}
+            onKeyDown={handleKey}
           >
             <InputTags.Tag.Text>{tag}</InputTags.Tag.Text>
             <InputTags.Tag.Close onClick={handleCloseTag} />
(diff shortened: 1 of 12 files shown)


ekrasnopolskaia:
InputTag doesn't have a11y tab, should we make up something?

website/docs/components/input-tags/examples/basic.jsx
```diff
@@ -50,14 +56,20 @@ const Demo = () => {
             editable
             data-id={idx}
             onClick={handleEditTag}
+            onKeyDown={handleKey}
           >
             <InputTags.Tag.Text>{tag}</InputTags.Tag.Text>
             <InputTags.Tag.Close onClick={handleCloseTag} />
```
lsroman, line 62: `onClick={handleCloseTag}` why is it here? It doesn't do anything

website/docs/components/input-tags/examples/basic.jsx
```diff
@@ -50,14 +56,20 @@ const Demo = () => {
             editable
             data-id={idx}
             onClick={handleEditTag}
+            onKeyDown={handleKey}
```
lsroman, line 59: Can do as with icons? What if was call a click then a keydown is called

website/docs/components/input-tags/examples/basic.jsx
```diff
@@ -50,14 +56,20 @@ const Demo = () => {
...
           </Tooltip.Trigger>
           <Tooltip.Popper>tag</Tooltip.Popper>
         </Tooltip>
       ))}
-      <InputTags.Value value={value} onChange={updateValue} onBlur={handleBlurInput} />
+      <InputTags.Value
+        value={value}
+        onChange={updateValue}
+        onBlur={handleBlurInput}
+        aria-label="input with tags"
```
lsroman, line 71: Could you synchronize with ux specialists and understand what is more important here, a placeholder or a label?

lsroman (changes requested):
I've added a few changes. Let's discuss these changes in a voice so it will be faster?

website/docs/components/input-tags/examples/basic.jsx
ekrasnopolskaia: Julia said that label is more important

website/docs/components/input-tags/examples/basic.jsx
phytonmk: Same question

ekrasnopolskaia: Why? It deletes the tag when you click the cross. Clicking on tag sends tag to edit, but clicking on the cross deletes it

lsroman (approved):
😍 thanks for your work!

PR #359, opened 2022-08-22
[input, radio, textarea, time-picker] added missing types

freeyoungstrong:


lsroman (changes requested):
It's so strange that we missed these important types. Thanks for fixing it!
Add a changelog and you can safely merge your fixes 😉

PR #374, opened 2022-09-02
[input-mask] a11y

freeyoungstrong:
## Description

Added additional `span` with `role="alert"` to inform user which numbers were entered.
Added `title` checking attribute for describing mask format.
Updated examples.

## Motivation and Context

Make input-mask component more accessible

## How has this been tested?

with VoiceOver app

## Types of changes

## Checklist:


freeyoungstrong:
great work! a11y works ok, and everything is ok. I just want to turn your mind to some points in style. first screen - after your changes and second screen - before your changes. I don't think it is problem, but dash in placeholder not grey
Screenshot 2022-09-21 at 23 22 35
Screenshot 2022-09-21 at 23 22 48
another point is that after entering 4 digits input sign is closer to left (4th position) than to right (5th position). first screen - after your changes and second screen - before your changes.
Screenshot 2022-09-21 at 23 27 39
Screenshot 2022-09-21 at 23 30 01

phytonmk:
Thank you for review!

I have fixed the issue with placeholder color. Position of cursor that are not pinned to the next masked symbol is not either bug or feature. It is just consequence of new input handling approach where non-user symbols (like a space between digit and underscore) are not added to input value when possible

PR #383, opened 2022-09-14
[chart, table] marked these packages as deprecated

freeyoungstrong:
## Description

Marked packages `@semcore/chart` and `@semcore/table` as deprecated and added recommendations to use `@semcore/d3-chart` and `@semcore/data-table` instead.

## Types of changes

## Checklist:

### DIFF
--- a/semcore/input-tags/src/InputTags.tsx
+++ b/semcore/input-tags/src/InputTags.tsx
@@ -17,9 +17,8 @@ export interface IInputTagsProps extends Omit<IInputProps, 'size'>, IScrollAreaP
    * @default m
    */
   size?: InputTagsSize;
-  /** Event is called when tag needs to be added
-   * @deprecated use `onAppend` instead
-   */
+  /** @deprecated use `onAppend` instead */
+  /** Event is called when tag needs to be added */
   onAdd?: (value: string, event: React.KeyboardEvent | React.ClipboardEvent) => void;
   /** Event is called when tags need to be added */
   onAppend?: (values: string[], event: React.KeyboardEvent | React.ClipboardEvent) => void;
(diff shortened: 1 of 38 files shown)


semcore/input-tags/src/InputTags.tsx
```diff
@@ -17,8 +17,9 @@ export interface IInputTagsProps extends Omit<IInputProps, 'size'>, IScrollAreaP
    * @default m
    */
   size?: InputTagsSize;
-  /** @deprecated use `onAppend` instead */
-  /** Event is called when tag needs to be added */
+  /** Event is called when tag needs to be added
+   * @deprecated use `onAppend` instead
```
phytonmk, line 21: Why this file was affected?

phytonmk (changes requested):
**I was waiting for that almost for twho years!**

But, two things:
1. `input-tags` was affected but should be not
2. Changelogs updates are also needed

semcore/input-tags/src/InputTags.tsx
freeyoungstrong: I just walked near this file and noticed that the description of the prop was written incorrectly I mean users don't see that it was deprecated and need to use another function
Screenshot 2022-09-15 at 18 30 04

freeyoungstrong: I just fixed description

freeyoungstrong: Screenshot 2022-09-15 at 18 31 28

freeyoungstrong: but yes I understand thta in this case commit message and these changes are not the same. anyway I can return everything as it was :)

PR #415, opened 2022-10-13
[data-table] Table in table with accordion and fixed column

lsroman:
## Description

There are two problems:
1. When using a table within a table, the calculation of column widths was out of sync. Therefore, I added one more argument from which you can take the size of the column.
2. When using a fixed column, the sticks do not work if one of the parents has an `overlay != visible`. I've added a few properties to block the overlay from being set by default. The accordion had it for animation. The table had it for scroll.

## Motivation and Context

New tables don't work with complex use cases. It blocks the transition to them.

## How has this been tested?

I added a few examples in documentation.
Open the site and show examples. For this, I deployed RC https://intergalactic-mj3h2dvptq-uk.a.run.app/ 
You can play in the playground 🧑‍💻

## Types of changes

## Checklist:


semcore/animation/src/Collapse.jsx
```diff
@@ -23,21 +23,25 @@ const style = sstyled.css`
...
+    }
     setHeightVar(e.currentTarget.scrollHeight + 'px');
   }, []);
 
   const handlerAnimationEnd = useCallback((e) => {
     if (e.currentTarget !== e.target) return;
     if (onAnimationEnd) onAnimationEnd(e);
-    e.currentTarget.style.overflow = 'initial';
+    if (overflowHidden) {
+      e.currentTarget.style.overflow = 'initial';
```
phytonmk, line 43: Maybe it should not use `initial` but take the previous value of `e.currentTarget.style.overflow` that may be stored before animation start?

semcore/data-table/src/DataTable.tsx
```diff
@@ -234,12 +236,14 @@ class RootDefinitionTable extends Component<AsProps> {
...
       const column = this.columns.find((column) => column.name === name);
-      columnsChildren.push({
+
+      const columnChildren = {
         get width() {
           return this.props.ref.current?.getBoundingClientRect().width || 0;
         },
         name,
-        cssVar: createCssVarForWidth(name),
+        varWidth: !varWidth || varWidth === 'inhered' ? createCssVarForWidth(name) : varWidth,
```
phytonmk, line 245: What is `inhered`? May be you meant `inherited` or `inherit`?

semcore/data-table/src/types.ts
```diff
@@ -22,7 +22,8 @@ export type Column<
   resizable?: boolean;
   sortable?: boolean | SortDirection;
   sortDirection: SortDirection;
-  cssVar: string | string[];
+  varWidth: string;
+  setVar: boolean;
```
phytonmk, line 26: Do we really need to expose so low-level apis?

semcore/data-table/src/DataTable.tsx
```diff
@@ -258,7 +261,13 @@ class RootDefinitionTable extends Component<AsProps> {
           children,
           ref: column?.props?.ref || React.createRef(),
         },
-      });
+      } as Column;
+
+      // @ts-ignore
```
phytonmk, line 266: Why do u use ts-ignore? 🤨

semcore/data-table/src/DataTable.tsx
lsroman: Yes, of course. I had a mind `inherit`

semcore/data-table/src/types.ts
lsroman: This is an internal api 
https://github.com/semrush/intergalactic/blob/9b89f4c926c1fe14021f3a7d5f19041a493c7dc0/semcore/data-table/src/DataTable.tsx#L157

semcore/data-table/src/DataTable.tsx
lsroman: Screenshot 2022-10-19 at 11 26 34
Without it, TS go crazy 🤯

phytonmk: Ts is not going crazy. You just provided wrong typings for `column` variable.

You have either initialize it with corresponding value (`let columns: Cloumn[] = []`) or provide corresponding typings (`let columns: Column[] | undefined`).

semcore/data-table/src/types.ts
phytonmk: As I see from [coming examples](https://github.com/semrush/intergalactic/pull/415/files/9b89f4c926c1fe14021f3a7d5f19041a493c7dc0#diff-b80726e2311033aa13b20cf627bf4fcf036a4d883e9211cbe459d6e60a9c4885R38-R41) these apis are intended to be available for users so it's kinda enough to perceive them as public ones

semcore/data-table/src/DataTable.tsx
lsroman: It became clearer. The error was not informative enough for me(

semcore/data-table/src/types.ts
lsroman: These are still internal types. And I just did not make public ones. Now I have [added](https://github.com/semrush/intergalactic/pull/415/files/9b89f4c926c1fe14021f3a7d5f19041a493c7dc0#diff-0fae13344b5ab0b2735577659c2fdeac367710308c912123bb21068a1cb63d2aR102) them.

PR #454, opened 2022-11-18
[website] updated illustrations and guides & added margin-bottom to P…

elenakhaas:


j-mnizhek (approved):
Everything is great!
I've only updated the snapshots of the ProductHead component through `pnpm test -- -u` comand and fixed a little bit text in the changelog

PR #456, opened 2022-11-21
Update/styles-for-tokens

j-mnizhek:
Small updates in styles to fit into our design tokens system:
- Indicator in Carousel
- Active item's border in ColorPicker
- State colors for Knob in Slider
- Background color for Sidebar in Wizard


j-mnizhek:
I am closing this commit due to Misha has already updated all styles according to the components in Figma

PR #462, opened 2022-11-30
[website] added A11y guides

elenakhaas:


website/docs/components/skeleton/skeleton-a11y.md
```diff
@@ -0,0 +1,20 @@
...
+title: A11y
+a11y: AA
+---
+
+@## Considerations for developers
+
+- When a change occurs in the UI, like a loading state, it's good to notify a user what is happening. While a skeleton loader visually indicates that a process is happening, it should also be announced by assistive technology through the use of `aria-live`.
+- See the WAI-ARIA `aria-live` [documentation](https://www.w3.org/TR/wai-aria-1.1/#aria-live) for more details.
+- Add the `screenreaderText` prop and add a message for assistive technology (Example: `screenreaderText=”Loading contents”`.
+- If multiple skeletons exist on a screen, you likely only need to add the `screenreaderText` prop to one of them.
```
j-mnizhek, line 11: Do we have such prop in our component?

website/docs/style/icon/icon-a11y.md
```diff
@@ -0,0 +1,49 @@
+---
```
j-mnizhek, line 1: This guide should be rewritten, because there is no such role as icon, and our Icon component doesn't have all this roles and attributes you've added to the guide.

j-mnizhek (changes requested):
Added some comments, and also see DM in Slack for the detailed review

website/docs/style/illustration/illustration-a11y.md
```diff
@@ -0,0 +1,53 @@
+---
+title: A11y
+a11y: AA
+---
+
+@## What component has
+
+### Roles & attributes
```
j-mnizhek, line 8: Do not sure if the component has all these attributes. Need to check.

website/docs/style/icon/icon-a11y.md
```diff
@@ -0,0 +1,43 @@
...
+a11y: AA
+---
+
+@## What component has
+
+### Roles & attributes
+
+The list below describes roles and attributes that component already has.
+
+| Role  | Attribute            | Element | Usage                                                                                                                        |
```
j-mnizhek, line 12: There is no role="img" and such atributes in the component yet.

You can read https://www.w3.org/TR/SVG-access/ and https://css-tricks.com/accessible-svgs/ to understand what we can add and what our developers should add.

j-mnizhek (changes requested):
There is no role="img" and such atributes in the Illustration component yet.

You can read https://www.w3.org/TR/SVG-access/ and https://css-tricks.com/accessible-svgs/ to understand what we can add and what our developers should add.

website/docs/style/icon/icon-a11y.md
j-mnizhek: And always check how component is read by screenreader. It usually helps to find all inconsistencies.

j-mnizhek: Updated guide for Icon component.

PR #506, opened 2023-01-12
[popper] when hover - does not move focus

freeyoungstrong:
## Description

Changed behaviour - now when hovering over a tooltip, the focus doesn't move to the tooltip.


phytonmk:
Better resolved in https://github.com/semrush/intergalactic/pull/509

PR #507, opened 2023-01-13
[notice-bubble] removed minimal height limitation

phytonmk:
## Description

I have removed `min-height` prop of notice bubble component.  

## Motivation and Context

I suppose `min-height` prop was added by mistake. Some of our users overwrite it with `@semcore/babel-plugin-react-semcore` and after removing it component looks nicer a bit.

## How has this been tested?

I have updated screenshot tests and don't see any edge case that might be broken.

## Screenshots:

Before:

After:

## Types of changes

## Checklist:


freeyoungstrong (approved):

lsroman:
Screenshot 2023-01-16 at 14 58 12
This was done to make it look better if there is no content.
Screenshot 2023-01-16 at 14 59 12
I understand that this is a stupid case) Can we just lower this height?

PR #519, opened 2023-01-23
[babel-plugin-react-semcore] removed css-purge package

freeyoungstrong:
## Description

Removed `css-purge` package due to build errors and lack of support from maintainers.
And added `postcss-combine-duplicated-selectors` instead of removed one.

screenshot of error: 
Screenshot 2023-01-23 at 18 58 27


tools/babel-plugin-styles/src/postcss.js
```diff
@@ -51,6 +52,7 @@ module.exports = function (options) {
       ),
     ),
     postcssHoverMediaFeature(),
+    combineSelectors(),
```
lsroman, line 55: Why is it included in this package?

lsroman (changes requested):
I did not understand why the plugin that removes duplicates was removed from "tbabel-plugin-react-semcore" and added to "tbabel-plugin-style""?
Why were the tests removed? And no new ones have been added?

PR #524, opened 2023-01-30
[tokens] update hex values for Facebbok, LinkedIn and Twitter tokens

j-mnizhek:


j-mnizhek:
Added in https://github.com/semrush/intergalactic/pull/528, so this can be closed with a light heart 😊

PR #526, opened 2023-02-06
[table] fix colgroup calculation

bbutkov:
For a table with one header level, if colSpan > 1, the column width is not calculated correctly

## Description

Added an extra pass through the first level of the header to set the width for elements with undefined width

## Motivation and Context

To fix this misbehavior

## How has this been tested?

I checked on my tables that this fix works for different types of tables and checked **local website** version of this fix.

I also have a code where I completely rewrote **colgroup** calculation function, but due to the lack of tests, I decided to leave only an additional pass through elements, this is much safer

**From**:
from

**To**:
to

## Types of changes

## Checklist:

# If appropriate sections above are not fulfilled properly do not even review my pull request and never approve it! Only after fulfilling sections above I will also remove this note.


freeyoungstrong (approved):
looks good, thanks 👍

PR #527, opened 2023-02-06
[card] fixed display Card.Title with Ellipsis component

freeyoungstrong:
## Description

Here is example of incorrect display `Card.Title` with `Ellipsis` component - https://codesandbox.io/s/blue-mountain-22k2rj?file=/src/App.js. It was fixed by changing `inline-flex` to `flex`. We still need `SWrapper` because of centered display of title of card and hint icon.
Now it looks like: 
Screenshot 2023-02-06 at 18 18 51

## Types of changes

## Checklist:


phytonmk (changes requested):
As was discussed changing block model of component is a breaking change.

PR #528, opened 2023-02-07
feature/add semantic radii tokens

j-mnizhek:
- Added semantic tokens for border-radius. They all refer to the base border-radius tokens.
- Fixed strange (I guess, old) border-radius styles for Dot and ScrollArea.
- Fixed styles for Divider and Breadcrumbs. Divider with `invert` theme had wrong styles before.


j-mnizhek (commented):
Yeah!

PR #545, opened 2023-02-20
[global] fix transform-runtime helpers import

bbutkov:
Fixing the problem of incorrect helper imports

## Description + Motivation and Context

Due to a bug in babel-runtime, not all helpers are imported correctly, instead, some of them are simply inlined INTO EVERY module, which affects code size and performance degradation

These are links to a detailed description of the problem

https://github.com/babel/babel/issues/10334
https://github.com/babel/babel/issues/10261
https://github.com/facebook/docusaurus/pull/2111

This fix will definitely improve code size and performance, but to make it even better, you need to get rid of **babel-runtime-helper** and use **@babel/preset-env** with properly configured **browserslist**, then native method will be used for the same code spread

## How has this been tested?

All tests passed, build does not crash, website works locally

**Before:**
before-core

**After:**
after-core

## Types of changes

## Checklist:

### DIFF
--- a/tools/babel-preset-ui/.babelrc.js
+++ b/tools/babel-preset-ui/.babelrc.js
@@ -11,7 +11,11 @@ module.exports = function (babel, opts = {}) {
       '@babel/plugin-proposal-export-default-from',
       '@babel/plugin-proposal-class-properties',
       'babel-plugin-preval',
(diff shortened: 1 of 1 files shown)


lsroman:
Thank you very much for work you have done!

PR #560, opened 2023-03-01
[table] fix colgroup calculation for StickyHead

bbutkov:
Use getBoundingClientRect().width instead of offsetWidth

## Motivation and Context
getBoundingClientRect().width is a **float** while offsetWidth contains an **integer**

Сell borders in table body and header may not match, since the sizes in the header are calculated in the function using offsetWidth, and the table body itself is dynamic and can be a float

Previously it was imperceptible, there were no vertical borders in the design tables

## How has this been tested?
Tested on my internal tables, also on local website and successful tests in ci

**Before:**
Снимок экрана 2023-02-28 в 20 06 10

**After:**
Снимок экрана 2023-02-28 в 20 06 40

## Types of changes

## Checklist:

### DIFF
--- a/semcore/table/CHANGELOG.md
+++ b/semcore/table/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+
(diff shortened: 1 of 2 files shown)


freeyoungstrong (approved):
LGTM. thanks a lot!

PR #575, opened 2023-03-10
WIP:[data-table,table] Change layout sort icon

lsroman:
## Description

By default, sorting runs over the text in any of its directions and at any column size. The place for the icon is not reserved, so the text is not pushed anywhere.
I also added line-height for table head as it was in the design. Because of this, a lot of screenshots have changed.

## Motivation and Context

To reduce the column size

## How has this been tested?

View screenshot tests

## Types of changes

## Checklist:


semcore/data-table/src/Head.tsx
```diff
@@ -92,7 +92,8 @@ class Head extends Component<AsProps> {
         fixed={column.fixed}
         resizable={column.resizable}
         sortable={column.sortable}
-        sortIconFloat={column.props.justifyContent?.includes('end')}
+        // sortIconReserv={column.props.justifyContent?.includes('center')}
```
freeyoungstrong, line 95: do we really need this comment?

semcore/data-table/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [3.8.0] - 2023-03-10
+
+### Changed
+
+- Changed the presentation of the sort icon. Now she always runs into the text.
```
freeyoungstrong, line 9: I think `it` will be more correct instead of `she`

semcore/table/src/style/table.shadow.css
```diff
@@ -106,50 +106,33 @@ SCellHead[sorting] {
     background-color: var(--intergalactic-table-th-primary-cell-hover, #e0e1e9);
   }
 
-  &:hover SSortWrapper {
-    display: block;
+  &:not([active]):hover SSortWrapper {
+    display: inline-flex;
+    margin-left: -16px;
```
freeyoungstrong, line 111: I think here should be design token value

semcore/table/src/style/table.shadow.css
```diff
@@ -106,50 +106,33 @@ SCellHead[sorting] {
...
-SSortWrapper:before {
+SSortWrapper::before {
   content: '';
   position: absolute;
-  top: -2px;
+  top: 0;
   right: 0;
-  width: 20px;
-  height: 16px;
+  width: calc(var(--intergalactic-spacing-1x, 4px) + 16px);
```
freeyoungstrong, line 134: and here shouldn't be design token value :sweat_smile: 
design tokens are not used for `width` and `height`

PR #590, opened 2023-03-20
[website] added Proximity principle guide

elenakhaas:


website/docs/patterns/proximity-principle/proximity-principle.md
```diff
@@ -0,0 +1,69 @@
...
+![](static/simplicity.png)
+
+### Continuation
+
+When there is an intersection between two or more objects, people tend to perceive them as a single uninterrupted object.
+
+![](static/continuation.png)
+
+
+In a text block, the distance between words is internal and between lines is external. Hence, according to the proximity principle, the interline spacing should be greater than the distance between words.
```
j-mnizhek, line 45: There is not enough division into sections in the guide. After the general principles, specific cases immediately follow. It is necessary to separate it all.

website/docs/patterns/proximity-principle/proximity-principle.md
```diff
@@ -0,0 +1,69 @@
...
+
+![](static/interline-spacing-yes-no.png)
+
+If a text block has a title, the distance between the title and the block should be greater than the line spacing. This rule is really easy. The external distance is always bigger than the internal.
```
j-mnizhek, line 49: I think it's worth giving a title to each case.

pnpm-lock.yaml
j-mnizhek, line 1: This file shouldn't be in the commit.

j-mnizhek (changes requested):
I left some comments. Need to update the guide according to them.

j-mnizhek:
I am closing this PR due to irrelevance. Maybe someday we will add such a guide to the DS, but this version isn't good unfortunately for bringing it to the master branch.

PR #595, opened 2023-03-23
[d3-chart] Added `additionalFields` in `a11yAltTextConfig` for extra …

ekrasnopolskaia:
…text description to the data when using a screen reader

## Description

Sometimes you may want to add a textual description to the screen reader data. for this purpose a new property has been added

## Types of changes

## Checklist:

### DIFF
--- a/semcore/d3-chart/src/a11y/hints.ts
+++ b/semcore/d3-chart/src/a11y/hints.ts
@@ -189,5 +189,5 @@ export const makeDataSummarizationConfig = (
     notableDiff: undefined,
     ...(config?.movingAverage ?? {}),
   },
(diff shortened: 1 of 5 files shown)


semcore/d3-chart/src/a11y/hints.ts
```diff
@@ -188,4 +189,5 @@ export const makeDataSummarizationConfig = (
     notableDiff: undefined,
     ...(config?.movingAverage ?? {}),
   },
+  additionalFields: [...config.additionalFields]
```
phytonmk, line 192: pls add optional chaining if config is not provided and ?? [] fallback if additionalFields is undefined

PR #600, opened 2023-03-27
[global] supported ignoring parent portals nesting

phytonmk:
## Description

I have added prop `ignorePortalsStacking` for portals.

## Motivation and Context

After animation update we faced following issue with tooltips in modals when tooltip is applied to elements which width is close to the width of the modal.

image

Going deeper showed that the reason is nested portals:  modal is rendered in portal, but creates portal provider inside of that. It ensured z-index are correct and position of inner poppers is related to position of modal (what is very important in animations!) on other hand in such situation popperjs modifier `computeStyles` finds parent node with scrolling and tries to place popper inside of that scrollable parent.

Such behavior is very good for most cases except cases when tooltip trigger size is almost equal to the size of scrollable parent, where popper is going to be placed. 

After implementing portals prop to disable stacking another issue was faced. When position of popper is not related to position of the animated container, position calculation is initiated on the first frame of animation and doesn't update later.

https://user-images.githubusercontent.com/31261408/227966292-34336837-d6ae-4b58-bd9e-b9268bd72367.mov

To deal with it I've added animation context – it allows popper in case of disabled portal stacking to listen to the animation start and end. When animation stars, popper is being updated every animation frame until animation ends (or timeout happens).

It doesn't work perfect because in cases of low performance browser will not give enough requested animation frame callbacks but in most cases it much increases UX while preserving working animations both from nested portals and from document root

https://user-images.githubusercontent.com/31261408/227967731-c39fdda2-8461-48ea-aa9d-0eb20d1d9d3e.mov 

## How has this been tested?

Only manual testing.

## Types of changes

## Checklist:

### DIFF
--- a/semcore/animation/CHANGELOG.md
+++ b/semcore/animation/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [1.10.0] - 2023-03-27
+
+### Added
+
+- Animation context that allows children components react to parent animation execution.
+
 ## [1.9.9] - 2023-03-28
 
 ### Changed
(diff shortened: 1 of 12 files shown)


ekrasnopolskaia (approved):
Looks great

PR #606, opened 2023-04-04
[d3-chart] fixed height calculating of StackBar and HorizontalBar components

freeyoungstrong:
## Description

Before there are problems with height of bars when several bars in one stack bar have minimum height. It was fixed by adding correct offset to each bar in stacked bars.

## How has this been tested?

snapshots

## Types of changes

## Checklist:


semcore/d3-chart/src/StackBar.jsx
```diff
@@ -18,6 +18,9 @@ class StackBarRoot extends Component {
     return { stack, r: 2 };
   };
 
+  offset = 0;
+  offsetArray = [0];
```
lsroman: Why is there already one value?) array must be empty in initial state

semcore/d3-chart/src/StackBar.jsx
```diff
@@ -52,6 +55,14 @@ class StackBarRoot extends Component {
       this.series.slice(seriesIndex + 1).some((bar) => bar[i][0] !== bar[i][1]) ? 0 : r,
     );
 
+    for (let i = 0; i < series.length; i++) {
+      if (series[i][1] - series[i][0] !== 0 && series[i][1] - series[i][0] < 0.1) {
```
lsroman, line 59: It's not correct to use constant 0.1

lsroman: You need `absHeight` from Bar

PR #608, opened 2023-04-11
update/table tokens

j-mnizhek:
- Added lighten modifier to two table cell tokens, to remove opacity from their bg.
- Removed website tokens.


semcore/utils/theme/light.json
```diff
@@ -950,14 +950,6 @@
           "description": "Active (selected) state of the inverted version of the tertiary control."
         }
       },
-      "lift-me-up": {
```
freeyoungstrong, line 953: I'm not sure, but probably this shouldn't be deleted

semcore/utils/theme/light.json
```diff
@@ -950,14 +950,6 @@
...
-        "value": "{gray.100}",
-        "type": "color"
-      },
-      "lift-me-up-hover-active": {
```
freeyoungstrong, line 957: I'm not sure, but probably this shouldn't be deleted too

PR #632, opened 2023-05-02
[badge] update default colors for bg in playground

j-mnizhek:
Updated default colors for Badge's background:
- `gray-300` to `gray-400`
- `orange-300` to `orange-400`
- `red-300` to `red-400`
- `blue-300` to `blue-400`
- `green-300` to `green-400`

From my side I cannot update the default bg color [right in the Badge component](https://github.com/semrush/intergalactic/blob/2c6dd4504759c3b46c3ba50ac1a90ca21154ebdd/semcore/badge/src/Badge.jsx#L13), got errors. So will add this to the task for updates.


j-mnizhek:
Wiil merge this after the default bg color for the Badge will be updated.

PR #649, opened 2023-05-31
[inline-input] fixed in some cases `onBlurBehavior` effect was called simultaneously with explicitly called confirm or cancel effect

phytonmk:


phytonmk:
Also need to test with keyboard

PR #677, opened 2023-07-17
Update/notice bubble guide

j-mnizhek:
- Update and fix NoticeBubble guide
- Remove recommendation about `font-smoothing` in the Typography guide according accessibility recommendations, and after the request from the core unit designer.


b0g3r:
@j-mnizhek FYI: We had in the past search requests with "font-smoothing" query and clicks on the Typography article. 

wdyt about adding notice about `font-smoothing: antialiased` into a11y tab?

j-mnizhek:
This setting affects accessibility, and together with the Coffee team we decided to remove this recommendation, and they will also remove it in their templates (or have already removed it).

I think I'll add some information to the accessibility tab later.

b0g3r:
Thanks Julia, the search request was indeed from someone on the Coffee team, so you can skip my message if you think it's no longer relevant 🙌

PR #808, opened 2023-10-10
[popper] Prevent to return focus to `Trigger` after popper is closed

ilyabrower:
## Motivation and Context
disable scrolling to a Select component when Select.Menu is opened and I click outside of the menu

## How has this been tested?
I have added unit tests, and I have tested it manually on playground

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/popper/__tests__/index.test.jsx
+++ b/semcore/popper/__tests__/index.test.jsx
@@ -302,4 +302,46 @@ describe('focus control', () => {
     expect(document.activeElement.nodeName).toBe('DIV');
     vi.useRealTimers();
   });
+
+  test('prevent focus return', () => {
+    let hidePopper = undefined;
+    vi.useFakeTimers();
+    const Component = () => {
+      const [visible, setVisible] = React.useState(true);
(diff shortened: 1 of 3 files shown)


semcore/popper/__tests__/index.test.jsx
```diff
@@ -302,4 +302,46 @@ describe('focus control', () => {
     expect(document.activeElement.nodeName).toBe('DIV');
     vi.useRealTimers();
   });
+
+  test('prevent focus return', () => {
```
msereniti, line 306: Basically I am very surprised that without `preventFocusCatch` functionality this test is passing 🤔

msereniti (changes requested):
The test should be updated/removed and the changelog note should correspond to the runtime update

semcore/popper/__tests__/index.test.jsx
ilyabrower: removed test, we don't need it

msereniti (approved):

Issue #747, opened 2023-09-12
Resolve color util should resolve design tokens


msereniti:
A lot of our components allows to pass color in `color` or `theme` prop. The prop accepts color as is or, as a name of color from hardcoded palette:

```js
const colors = preval`
  const path = require('path');
  module.exports = require('@semcore/babel-plugin-react-semcore').getColorVars(path.resolve(__dirname, './style/var.css'));
`;
```

All colors resolving goes via `resolveColor` function. We need to replace it with some kind of `useResolveColor` hook that will resolve colors by css variables + hardcoded default values.

PR #811, opened 2023-10-10
Resolvable design tokens

msereniti:
https://github.com/semrush/intergalactic/issues/747

The `useResolveColor` hook was added to utils as a replacement to `resolveColor` function. Instead of converting color key to it's hex value, it acts differently:
1. If the passed color is taken from the old palette – it sends (only in dev mode) warning into console and returns color as `resolveColor` do. 
2. If passed color is a design token (`--intergalctic-` prefix is optional). It takes it's default value from default theme file.
3. After getting default value, hook returns css variable expression with default value that may be used in css.

I made resolving to css variables because it allows absolutely smooth theme switching as it must work.

With this hook it's not not possible to apply color filters from js as we used to do in `semcore/utils/src/color.ts` file. So I changed component that was relying on that mechanism to use only css filters and properties to change the colors. For example, in `Tag` component the text color is also applied to background but the background also gets `opacity: 0.1; filter: brightness(1.5) saturate(1.5);` to make the text look nice on it. Same applied to badge and counter.

Also, that changes made it possible to use `--intergalactic-chart-palette-order-{n}` tokens in charts.  So I removed most uses of explicit `color` prop in charts documentation.

I had a lot of troubles with making automatic colors work in the charts' tooltip because due to some very weird hacks in the `semcore/core` the `getDotProps` on tooltip wasn't working. I have added additional context that seems to work nice but absolutely ugly in general.

The PR is ready for review.


semcore/link/src/style/link.shadow.css
```diff
@@ -52,7 +54,8 @@ SLink[keyboardFocused]::after {
 SLink[disabled] {
   opacity: var(--intergalactic-disabled-opacity, 0.3);
   cursor: default;
-  pointer-events: none; /* Disable link interactions */
+  pointer-events: none;
+  /* Disable link interactions */
```
ilyabrower, line 58: maybe stay on same line or before pointer-events?

semcore/utils/src/use/useColorResolver.ts
```diff
@@ -0,0 +1,206 @@
...
+  '--linkedin': '#1a7ab2',
+};
+
+const makeDeprecationMessage = (color: string) =>
+  `You are using ${color} color in your app. It's deprecated and will be removed from palette in the next major release. Please use colors from https://developer.semrush.com/intergalactic/style/design-tokens/ instead.`;
+
+export const useColorResolver = () => {
+  const contextTheme = useContextTokens();
+
+  return (color?: string) => {
```
ilyabrower, line 168: why do we use optional color here?

website/docs/data-display/donut-chart/examples/legend.tsx
```diff
@@ -53,11 +46,11 @@ export default () => {
       <Card.Header pt={4}> Chart legend</Card.Header>
       <Card.Body tag={Flex} direction='column'>
         <Flex flexWrap w={width}>
-          {piesList.map((pie) => {
+          {piesList.map((pie, index) => {
             return (
               <Checkbox
                 key={pie}
-                theme={pieColors[pie]}
+                theme={`chart-palette-order-${index + 1}`}
```
ilyabrower, line 53: as i see, we don't need to pass theme prop for now

semcore/base-trigger/src/style/link-trigger.shadow.css
j-mnizhek, line 1: Can we use tokens there? We have a special token for link hover and active states:
`--intergalactic-text-link-hover-active`

semcore/counter/__tests__/__image_snapshots__/counter-should-support-theme-props.png
j-mnizhek, line 1: Counter with colored background should not look like that. Black text on the orange and red backround is not readable and have bad contrast against the background. Can we use tokens in styles instead of filters?

image
image

semcore/link/src/style/link.shadow.css
```diff
@@ -88,9 +91,14 @@ SText {
...
-  &:hover {
-    color: var(--colorHoverText);
+SLink[text-color] {
+  color: var(--text-color);
+
+  &[active],
+  &:hover,
+  &:active {
+    color: var(--text-color);
+    filter: brightness(0.8);
```
j-mnizhek, line 101: Can we use token for the Link here instead of filter? Token called `--intergalactic-text-link-hover-active`. It's needed to be consistent with the design, and to simplify palette developing for the other themes (for example, dark).

website/docs/components/counter/examples/counter-textarea.tsx
j-mnizhek, line 1: Tokens for large text should not be used for texts smaller than 20px. Let's change token to `--text-secondary` here.

website/docs/components/counter/examples/counter-filter.tsx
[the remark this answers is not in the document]
j-mnizhek: OMG, now I understand why you changed it here. It's because `theme` prop here means color for the Counter.

website/docs/components/counter/examples/limit.tsx
j-mnizhek, line 1: I'll make changes to the right tokens for the example and will push it to the branch today.

website/docs/components/fullscreen-modal/examples/two-section.tsx
j-mnizhek, line 1: I will remove the unnecessary color prop here in my commit. Button with theme='muted' and use='tertiary' has the exact same color of the text by default.

website/docs/components/input/examples/placeholder.tsx
j-mnizhek, line 1: Will fix these to `text-secondary`.

website/docs/components/tab-line/components/playground.jsx
j-mnizhek, line 1: I will update it to `green-400` as it was in the design.

website/docs/components/tab-panel/components/playground.jsx
j-mnizhek, line 1: Will correct it the same way as the example for TabLine.

semcore/link/src/style/link.shadow.css
msereniti: Sure, it's just auto-formatting, will fix it,

semcore/utils/src/use/useColorResolver.ts
msereniti: For typings backward capability

semcore/base-trigger/src/style/link-trigger.shadow.css
msereniti: Sure, it's how it already made

j-mnizhek (approved):
Like the results!

PR #957, opened 2023-12-07
[website] update dark json for website

j-mnizhek:
Fixed some old tokens for DatePicker in dark.json. Commited with `--no-verify` because theme doesn't want to process 🤷🏻‍♀️


j-mnizhek:
Moved changes to https://github.com/semrush/intergalactic/pull/1292

Issue #785, opened 2023-09-29
Charts colorblindness support


msereniti:
**Request**

For greater accessibility of our chart library, it is important that the data on the graphs can be distinguished not only by color but also by non-color elements, so that one dataset can be distinguished from another. Similar remarks were made in the recommendations from Level Access:

> The blue and green lines and areas do not have sufficient contrast with each other and with the white background. They must have at least a 3:1 contrast ratio.

**What needs to be done:**

- [x] Prepare the technical implementation of this feature
- [ ] Select a set of SVG patterns for charts in the design
- [ ] Determine the order of using patterns on the charts
- [ ] Discuss the final implementation
- [ ] Implement the capability to fill charts with colored patterns in our chart library

PR #967, opened 2023-12-13
Charts patterns

msereniti:
## Motivation and Context

https://github.com/semrush/intergalactic/issues/785

For better a11y, our charts need opt-in color independent patterns drawing.

For now, patterns use is not very customizable and only two default patterns provided. This PR is opened to allow others to preview the feature and give their feedback.

## How has this been tested?

I will add a lot of VR tests later

## Screenshots:

Screen Shot 2023-12-13 at 12 57 46

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/d3-chart/src/Dots.jsx
```diff
@@ -76,26 +78,72 @@ function Dots(props) {
...
+            active={active}
+            hide={hide}
+            transparent={transparent}
+            cx={d3.x()(d)}
+            cy={d3.y()(d)}
+            r={radius}
+          />,
+        ),
+      );
+    } else {
```
ilyabrower, line 126: maybe it is better to define render as const before `acc.push` and don't use if-else ?

and why do you use `cx`, `cy`, `r` in first Dot and `x`, `y`, `radius` in second. Could we use the same props?

semcore/d3-chart/src/component/ChartLegend/LegendItem/LegendItem.type.ts
```diff
@@ -34,6 +35,10 @@ export type LegendItem = {
    * Additional info with sub-label or count
    */
   additionalInfo?: { label: string } | { count: number };
+
+  patternKey?: string;
+
+  patterns?: PatternsConfig;
```
ilyabrower, line 41: Let's add comments for new pops

semcore/d3-chart/src/Pattern.tsx
```diff
@@ -0,0 +1,645 @@
...
+  },
+} satisfies Record<string, Pattern>;
+
+const defaultPatternsList = Object.values(defaultPatterns);
+const patternsSync = new WeakMap();
+patternsSync.set(defaultPatternsList, []);
+
+export type PatternsConfig = true | keyof typeof defaultPatterns | Pattern | Pattern[];
+
+const getPatternByKey = (patternKey: string, patternsConfig: PatternsConfig = true) => {
```
ilyabrower, line 549: let's add here return type. Am I right, that it should be Pattern?

semcore/d3-chart/src/Pattern.tsx
```diff
@@ -0,0 +1,645 @@
...
+  if (!patternsSync.has(patterns)) {
+    patternsSync.set(patterns, []);
+  }
+  const patternsSyncList = patternsSync.get(patterns);
+  let index = patternsSyncList.indexOf(patternKey);
+  if (index === -1) {
+    index = patternsSyncList.length;
+    patternsSyncList.push(patternKey);
+  }
+  const patternIndex = index % patterns.length;
```
ilyabrower, line 572: looks like to difficult method... and it is not `getSomething`. It is get and sometimes store and magic =)
I suggest:
1) let's rewrite this place
```
const patternsSyncList = patternsSync.get(patterns);
const index = patternsSyncList.indexOf(patternKey);
if (index === -1) {
    patternsSyncList.push(patternKey);
    return patterns[0];
}
return patterns[index];
```
This is optional - just for think about it.
2) Maybe it is better to make a patternsList (default) some object (Map) with method to add some new patterns in it? And it will be obvious manually action - addSomeNewPattern. And getter will be just getter.

website/docs/data-display/area-chart/area-chart-d3-code.md
```diff
@@ -401,7 +403,7 @@ const Demo = () => {
 
   return (
     <>
-      <ChartLegend
+      <ChartLege
```
j-mnizhek, line 406: Wow, what's going on here. I'll fix this right now

semcore/d3-chart/src/Dots.jsx
msereniti: > and why do you use cx, cy, r in first Dot and x, y, radius in second. Could we use the same props?

It's the reason why I made if-else instead of single statement with different render prop. 
By default the dot is a circle that may change it's radius. It needs to use the `cx` and `cy` props to center correctly.

In the other case we have PatternSymbol that is a nested svg images. Centering all thing except the circle is a hard process in the svg world. So, it needs `x` and `y` props only and also need to compute width and height offset to render in the right place.

semcore/d3-chart/src/Pattern.tsx
msereniti: Yes, you are right

semcore/d3-chart/src/Pattern.tsx
msereniti: > let's rewrite this place

Ok, that makes sense that this attempts was too messy. 
1. I renamed the function to `resolvePattern`.
2. I wrapped all the things around the `WeakMap` into function called `resolvePatternIndex`.
3. The `WeakMap` instead of _syncingThing_  is not called `patternsKeysOrderCache` to make it more clear why it exists.

> Maybe it is better to make a patternsList (default) some object (Map) with method to add some new patterns in it? And it will be obvious manually action - addSomeNewPattern. And getter will be just getter.

That was how the first approach was made. But actually this has not so much sense. 

Default patterns list allows developers enable patterns by just passing `patterns={true}` our of the box. If develops want to set custom pattern, most likely that they want to set it on specific parts of the chart instead of adding it to the list that is used for automatic patterns distributing and may affect other charts behavior.

ilyabrower (approved):

PR #973, opened 2023-12-14
[website] added basic blog functionality

msereniti:
Also need to write first posts :)


j-mnizhek:
I have few comments;

1. I'd recommend to divide first post into two parts and add two headings for the simplier readability. For example, the first part can be titled as "Reasons", the last one "How to migrate" or something like this.

2. For blog posts we now have two dates. Maybe we should leave one?

Screenshot 2024-02-09 at 19 20 21

3. We need to fix blog section styles on the main page a little bit. Now h2 heading "Last updates" is too small. We can inherit the same styles for this heading as we have in the documentation.
image

4. Can we use the same styles for the article as we have for the cards above on the main page?

PR #980, opened 2023-12-19
[input-mask] fixed that component was breaking a page when initial value was not allowed by outer pipe

msereniti:
## Motivation and Context

It was found that if initial value is not allowed by outer pipe function, component blows up due to error inside of `input-mask-core`. Input mask core have no updated last 5 years and migrating to some other library is very painful (but needed in the future).

Here is an example when it was initially found – selected day is disabled https://codesandbox.io/p/sandbox/frosty-cloud-25yvw7?file=%2Fsrc%2FApp.tsx

Also you can checkout added test to for more simple example.

## How has this been tested?

Test that was added wasn't initially was throwing error like 'unable to read xxx.length of undefined` from somewhere of compressed `text-mask-core`.

After change it passed well. DatePicker example from above also works fine after the fix.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/d3-chart/CHANGELOG.md
+++ b/semcore/d3-chart/CHANGELOG.md
@@ -6,299 +6,299 @@ CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangel
 
 ### Fixed
 
-* Squeeze of shapes in `ChartLegendTable`.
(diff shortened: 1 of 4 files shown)


ilyabrower (approved):
You don't need this, I think: `semcore/d3-chart/CHANGELOG.md`
Approve

PR #1001, opened 2024-01-08
Update/accordion examples

j-mnizhek:
We had some visual issues with Accordion examples in the documentation. I've fixed examples' styles a little bit, so they meet the design requirements.


msereniti (approved):
Woah, a little but amazing change

PR #1042, opened 2024-01-19
[outside-click, utils] fixed outside clicks handling in shadow dom

msereniti:
## Motivation and Context

We recently moved all website examples into Shadow DON while all components was never tested to work in Shadow DOM. It was found that https://developer.semrush.com/intergalactic/utils/popper/popper render function examples closes immediately after opening. 

The example was broken because OutsideClick tries to get what element is clicked and in case of using shadow dom elements it's the shadow dom container – the whole playground. So, for outside click component every click was the outside one.

## How has this been tested?

The change is extremely simple – I've just made it to support shadow DOM with full backward capability. 
So I think we do not need shadow DOM tests for now. 

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/outside-click/src/OutsideClick.tsx
+++ b/semcore/outside-click/src/OutsideClick.tsx
@@ -31,15 +31,17 @@ const noop = () => {};
 function OutsideClick(props: IFunctionProps<IOutsideClickProps>) {
   const { Children, forwardRef, root, excludeRefs = [], onOutsideClick = noop } = props;
   const children = getOriginChildren(Children);
-  const nodeRef = React.useRef(null);
-  const targetRef = React.useRef(null);
+  const nodeRef = React.useRef<Node | null>(null);
+  const targetRef = React.useRef<Node | null>(null);
 
   const handleRef = useForkRef(children ? children.ref : null, nodeRef, forwardRef!);
 
   const handleOutsideClick = useEventCallback((event: any) => {
     const isTargetEvent = [...(excludeRefs as any), nodeRef]
       .filter((node) => getNodeByRef(node))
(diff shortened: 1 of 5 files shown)


semcore/outside-click/src/OutsideClick.tsx
```diff
@@ -30,23 +31,25 @@ const noop = () => {};
...
+  const targetRef = React.useRef<Node | null>(null);
 
   const handleRef = useForkRef(children ? children.ref : null, nodeRef, forwardRef!);
 
   const handleOutsideClick = useEventCallback((event: any) => {
     const isTargetEvent = [...(excludeRefs as any), nodeRef]
       .filter((node) => getNodeByRef(node))
-      .some((node) => getNodeByRef(node)?.contains(targetRef.current || event.target));
+      .some((node) =>
+        getNodeByRef(node)?.contains(targetRef.current || (getEventTarget(event) as Node | null)),
```
ilyabrower, line 43: `as Node` is incorrect casting, as I see

semcore/utils/src/getEventTarget.ts
```diff
@@ -0,0 +1,11 @@
+export const getEventTarget = (event: React.SyntheticEvent | Event): EventTarget | null => {
+  if ('composedPath' in event) {
+    return event.composedPath()?.[0] || event.target;
+  }
+  if (event.nativeEvent) {
+    if ('composedPath' in event.nativeEvent) {
```
ilyabrower, line 6: maybe `if (event.nativeEvent && 'composedPath' in event.nativeEvent) {` ?

PR #1069, opened 2024-02-02
[86bx9tb3v][button] Fixed paddings for icon in button

ilyabrower:
## How has this been tested?
I've tested it manually and added unit test

BEFORE:
Screenshot 2024-02-02 at 11 42 38

AFTER:
Screenshot 2024-02-02 at 11 41 40

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/button/__tests__/__image_snapshots__/button-renders-correctly-with-one-addon-as-props-for-large-size.png
j-mnizhek, line 1: I think we need a test for a button with L size and icon with M size inside, because now button with M sized icon has not square form, but rectangular, which is not correct. Actually we can see this in the screenshots for buttons in DatePicker.

j-mnizhek (changes requested):
Maybe we should add min-width for the buttons since L sized button become rectangle instead of square in cases when addon is smaller than expected? 🤔

PR #1075, opened 2024-02-05
[86bx705bh] attempted to remove focus return elements

msereniti:
I've removed hidden focus return and catch elements. Instead I've added a timeout (2000ms) after popper close in which popper ignores trigger focus events.

It seems to work much better than the pervious approach but may produce unexpected bugs.

I've set timeout to 2000ms cause on the slow devices popper close time may really take a lot of time like 1s. On the other hand it creates a possible scenario that may seem buggy – if user closes popper with mouse and then goes to trigger with keyboard in that 2000ms the popper will not appear on first focus (after blur and focus again it will). I suppose that this case will have no sense if we will open poppers by keyup/keydown on focused trigger.


semcore/popper/__tests__/index.test.jsx
```diff
@@ -213,6 +213,46 @@ describe('focus control', () => {
     expect(PopperElement).toHaveFocus();
   });
 
+  test('loop', async () => {
```
ilyabrower, line 216: maybe something like `should lock focus inside popper` ?

semcore/popper/src/Popper.jsx
```diff
@@ -269,6 +278,16 @@ class Popper extends Component {
         }
       }, 0);
     });
+    const ignoringDuration = 2000;
```
ilyabrower, line 281: I'm still disagree with timeout

semcore/utils/src/focus-lock/setFocus.ts
```diff
@@ -0,0 +1,76 @@
+import canUseDOM from '../canUseDOM';
+import { getFocusableIn } from './getFocusableIn';
+
+const unsafeSetFocus = (
+  node: HTMLElement,
+  previousActiveElement?: HTMLElement,
```
ilyabrower, line 6: great!) I don't understand we need it in `focus-lock` package as required prop

semcore/utils/src/focus-lock/setFocus.ts
```diff
@@ -0,0 +1,76 @@
...
+    { capture: true },
+  );
+  document.addEventListener(
+    'mousedown',
+    () => {
+      lastUserAction = Date.now();
+    },
+    { capture: true },
+  );
+}
```
ilyabrower, line 74: I'm not sure, that this part of code should be places like this, for calling every time we use this file

msereniti: It must be. We should listen to this event regardless to any focus lock consumer be mounted or not. Other way we can't provide "safe" focus set because focus set call may be called at any time.

Anyway, this code was existing before (but now it has `capture: true`).

ilyabrower (approved):

PR #1084, opened 2024-02-08
[86bxdefw0][utils] fixed Logic of returning focus to the trigger component if trapRef has been removed from DOM.

ilyabrower:
## Motivation and Context
We have a problem, that after unmount `Modal` in some parent component, `returnFocus` doesn't work correct. It was because in it we expect, that we have a ref to a dom node, but we don't, so, I've changed logic of this method.

## How has this been tested?
I've tested it manually and added unit test

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/utils/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [4.20.2] - 2024-02-08
+
+### Fixed
+
+- Logic of returning focus to the trigger component if `trapRef` has been removed from DOM.
```
msereniti, line 9: Probably it would be better to say that `Fixed edge case bug when browser focus should be returned from modal, but a button that triggered modal opening is already unmounted`.

msereniti (approved):

PR #1088, opened 2024-02-09
[86bxdgxqe][website] migrated docs to intergalactic package use, added docs and import package switcher

msereniti:
Screen Shot 2024-02-09 at 18 14 53
Screen Shot 2024-02-09 at 18 15 02


j-mnizhek:
Noticed that toggle is shown not for all sections, only for pages inside the Components & Patterns section. If I, for example, go to pages in Foundation section, it disappears 👀 
image

Shall we leave this behavior? I think it's better to show it for all documentation, because from the users side it's not clear why for some pages it's hidden, and for some aren't.

j-mnizhek:
Michael, great update! 😍

Let's add small nice improves for styles:
1. Hover styles for switcher option which are not highlighted. It will help understand that the option can be clicked
image

2. Hover style for "About API Reference" link: underline text-decodation.
image

j-mnizhek (approved):

package.json
```diff
@@ -8,14 +8,14 @@
     "install-n-test": "pnpm install && pnpm test",
     "install-n-browser-test": "pnpm install && pnpm playwright test --config playwright.browser.config.ts",
     "test": "vitest",
-    "test:setup": "docker build -t intergalactic-playwright -f playwright.dockerfile .",
+    "test:setup": "docker build -t intergalactic-playwright-2 -f playwright.dockerfile .",
```
ilyabrower, line 11: why?

website/docs/components/breadcrumbs/breadcrumbs.md
```diff
@@ -61,7 +63,7 @@ Margins between the links inside the breadcrumbs is 8px.
 
 ## Long links truncation
 
-If there isn’t enough place for the breadcrumbs (screen is too small or the text is too long), the last item is collapsed into `ellipsis`.
+If there isn’t enough place for the breadcrumbs (screen is too small or the text is too long), the last item is collapsed into `ellipsis` .
```
ilyabrower, line 66: you have strange spaces before `.` or `,` somewhere. and sometimes after `;`

PR #1107, opened 2024-02-19
[86bxdetnp][d3-chart] fixed that a11y module "Skip to content after plot" link was not working in some cases

msereniti:
## Motivation and Context

`heavyFindNextFocusableElement` was made with a lot of excess logic and in a quite strange way. I much simplified it and ensured that it works in tested and in previously non-working cases.

## How has this been tested?

Manually & with unit tests.

## Types of changes

- [ ] Nice improve.

## Checklist:


ilyabrower:
What do you think about add possibility to pass a ref to an element on which we should move after press "Skip to content after plot" to the Plot as prop?

PR #1121, opened 2024-02-22
[typography, website] added List.Item.Content to api docs

ilyabrower:
I've just added List.Item.Content types to docs


msereniti:
You should add usage examples as well

msereniti (approved):

PR #1123, opened 2024-02-22
[86bxkev49][base-trigger] fixed that filter trigger was setting aria attributes to wrapper instead of inner trigger

msereniti:
## Motivation and Context

Our Filter trigger was found to not be accessible cause it was setting aria attributes on outer wrapper instead of inner trigger. I've fixed it. Solution is not perfect cause we need to add deep component customization but it's good enough just to make it accessible.

## How has this been tested?

Manual testing, soon I will enable a11y tests.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/base-trigger/src/FilterTrigger.jsx
+++ b/semcore/base-trigger/src/FilterTrigger.jsx
@@ -18,6 +18,17 @@ import { setFocus } from '@semcore/utils/lib/use/useFocusLock';
 
 import style from './style/filter-trigger.shadow.css';
 
+const filterTriggerInputProps = [
+  ...inputProps,
+  'aria-controls',
+  'aria-haspopup',
+  'aria-expanded',
+  'aria-disabled',
+  'aria-activedescendant',
+  'aria-label',
+  'aria-labelledby',
+];
+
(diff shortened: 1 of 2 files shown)


semcore/base-trigger/src/FilterTrigger.jsx
```diff
@@ -82,6 +93,13 @@ class RootFilterTrigger extends Component {
         aria-label={getI18nText('filter')}
         __excludeProps={['id']}
         use:role={role}
+        use:aria-controls={undefined}
+        use:aria-haspopup={undefined}
+        use:aria-expanded={undefined}
+        use:aria-disabled={undefined}
+        use:aria-activedescendant={undefined}
+        use:aria-label={undefined}
+        use:aria-labelledby={undefined}
```
ilyabrower, line 102: it look's too ugly)

ilyabrower (approved):

semcore/base-trigger/src/FilterTrigger.jsx
msereniti: Sorry...

PR #1124, opened 2024-02-22
[86bxfj0xv][time-picker] fixed horizontal offset

msereniti:
## Motivation and Context

`input` tags in time picker are very narrow. It's narrow because it needs background highlight when focused. This background highlight cannot be made by `:after` or `:before` pseudo elements because `input` tag cannot contain it. 

While Select trigger is more narrow than popper, it's displayed incorrectly. To overcome it, I've added default `offset` prop that compensate it on popper.

## How has this been tested?

Our screenshot tests can't screenshot popper in right position, so only manually.

## Screenshots:

Before:

Screenshot 2024-02-22 at 16 25 17

After:

Screenshot 2024-02-22 at 16 09 23

## Types of changes

- [ ] Nice improve.

## Checklist:

- [ ] I have updated the documentation accordingly or it's not required.

### DIFF
--- a/semcore/time-picker/src/PickerInput.jsx
+++ b/semcore/time-picker/src/PickerInput.jsx
@@ -46,7 +46,7 @@ class ItemPicker extends Component {
 
   state = {
     dirtyValue: undefined,
-    visible: true,
+    visible: false,
   };
 
   parseValueWithMinMax = (value) => {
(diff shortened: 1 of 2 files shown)


semcore/time-picker/src/PickerInput.jsx
```diff
@@ -43,7 +46,7 @@ class ItemPicker extends Component {
 
   state = {
     dirtyValue: undefined,
-    visible: false,
+    visible: true,
```
ilyabrower, line 49: why did you change default state value?

msereniti: By mistake, rolled it back

PR #1125, opened 2024-02-22
[86bxew1rk][d3-chart] `Line.Dots` and `Area.Dots` components `display` prop now accepts function that allows developers to granularly control which dots should be displayed

msereniti:
## Motivation and Context

`@semcore/chart` was supporting control over which dots might be rendered. So we need to support it in `@semcore/d3-chart`.

## How has this been tested?

Manually + new unit test.

## Screenshots:

See changed files.

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/d3-chart/src/types/Area.d.ts
```diff
@@ -30,8 +30,8 @@ export type AreaProps = Context & {
 /** @deprecated */
 export interface IAreaDotsProps extends AreaDotsProps, UnknownProperties {}
 export type AreaDotsProps = Context & {
-  /** Show all Dot */
-  display?: boolean;
+  /** Show dots */
+  display?: boolean | ((index: number, active: boolean, single: boolean) => boolean);
```
ilyabrower, line 34: I think, we should pass a value here too

semcore/d3-chart/src/types/Line.d.ts
```diff
@@ -32,8 +32,8 @@ export type LineProps = Context & {
 export interface ILineDotsProps extends LineDotsProps, UnknownProperties {}
 export type LineDotsProps = Context &
   Omit<FadeInOutProps, 'display'> & {
-    /** Show all Dot */
-    display?: boolean;
+    /** Show dots */
+    display?: boolean | ((index: number, active: boolean, single: boolean) => boolean);
```
ilyabrower, line 36: And here too

semcore/d3-chart/src/types/Area.d.ts
msereniti: I intentionally didn't passed it because we can't type it properly and index is totally enough to get value from initial data array

ilyabrower (approved):

PR #1138, opened 2024-02-27
[website] fixed chart palettes

ilyabrower:
replaced old tokens with new `intergalacitc-chart-palette` tokens


j-mnizhek (approved):
Great! Thank you ❤️

PR #1142, opened 2024-02-28
[86bxcxzxy][date-picker] a11y fixes

msereniti:
## Motivation and Context

This PR contains a set of critical a11y fixes of date picker components.

1. I have rewrote range comparator keyboard handler to handle popper keydowns.
2. I added `unitRefs` – it allows to focus on `CalendarUnit` when user navigates with arrows.`
3. In all date picker poppers I removed `event.target` check as it was preventing keyboard handling when calendar unit is focused.
4. I added `abbr` attribute on calendar grid for all "short" forms of date terms. The `propsForElement` awful regexp was not allowing to set it, so I modified the regexp.
5. I added aria-live region that produces description of what happens when user selects date ranges.
6. I made date-picker popper elements focusable and not aria-hidden. Also I changed focus order to focus on periods and apply/clear button first.
7. Also I found that in range comparators periods presets had no padding – i've added it.
8. I've removed `aria-expanded` attribute from input trigger as it was violating a11y automatic checks. To ensure with it, I've updated a11y test from using old button trigger and replaced it with input trigger.

## How has this been tested?

With existing and updated unit tests. 

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/date-picker/src/components/Calendar.jsx
```diff
@@ -184,14 +188,74 @@ class CalendarAbstract extends Component {
...
+    if (value.length === 1 || compare.length === 1) {
+      description += t('selectingStarted') + '. ';
+    } else if (value.length === 2 || compare.length === 2) {
+      description += t('selectingFinished') + '. ';
+    }
+
+    if (actionsDescribing === 'range') {
+      if (value[0] || value[1]) description += t('dateRange') + ' ';
+      if (value[0]) description += t('fromDate', value[0]) + ', ';
+      if (value[1]) description += t('toDate', value[1]) + '. ';
```
ilyabrower, line 223: it will be a little bit strange description, if we haven't value[1]... no?

semcore/date-picker/src/components/Calendar.jsx
```diff
@@ -184,14 +188,74 @@ class CalendarAbstract extends Component {
...
+    } else if (actionsDescribing === 'range-compare') {
+      if (compare.length === 1) {
+        description += t('dateRange2') + ' ';
+        if (value[0]) description += t('fromDate', value[0]) + ', ';
+        if (value[1]) description += t('toDate', value[1]) + '. ';
```
ilyabrower, line 228: why do you use `value` instead of `compare`?

semcore/date-picker/src/components/Calendar.jsx
```diff
@@ -261,13 +325,17 @@ class CalendarDaysRoot extends CalendarAbstract {
...
 
     return sstyled(styles)(
-      <SCalendar render={Box} aria-hidden='true'>
+      <SCalendar render={Box}>
         <CalendarWeekDays locale={locale} />
         <SGridDays onMouseLeave={this.handleMouseLeave}>
           <Children />
         </SGridDays>
+        <span aria-live='polite'>
+          <ScreenReaderOnly>{description}</ScreenReaderOnly>
```
ilyabrower, line 337: maybe we should add `aria-live` to the `ScreenReaderOnly` as prop?

semcore/utils/src/propsForElement.ts
```diff
@@ -19,7 +19,7 @@ children dangerouslySetInnerHTML key ref autoFocus defaultValue valueLink defaul
...
...rCanc|lab)el|(?:allowFullScre|hidd)en|systemLanguage|(?:(?:o(?:nPointer(?:Ent|Ov)|rd)|allowReord|placehold|frameBord|paintOrd|post)e|repeatDu|d(?:efe|u))r|v(?:Mathematical|ert(?:Origin[XY]|AdvY)|alues|ocab)|(?:pointerEve|keyPoi)nts|(?:strokeLineca|onPointerU|itemPro|useMa|wra|loo)p|h(?:oriz(?:Origin|Adv)X|ttpEquiv)|(?:vI|i)deographic|unicodeRange|mathematical|vAlphabetic|u(?:nicodeBidi|[12])|(?:fontStretc|hig)h|(?:(?:mar(?:ker|gin)W|strokeW)id|azimu)th|(?:xmlnsXl|valueL)ink|mediaGroup|spellCheck|(?:text|m(?:in|ax))Length|(?:unitsPerE|optimu|fro)m|r(?:adioGroup|e(?:sults|f[XY]|l)|ows|[xy])|a(?:rabicForm|l(?:phabetic|t)|sync)|abbr|pathLength|innerHTML|xlinkShow|(?:xlinkHr|glyphR)ef|(?:tabInde|(?:sand|b)bo|viewBo)x|(?:(?:href|xml|src)La|kerni)ng|autoPlay|o(?:verflow|pen)|f(?:o(?:ntSize|rm?)|il(?:ter|l))|r(?:e(?:quired|sult|f))?|divisor|p(?:attern|oints)|unicode|d(?:efault|ata|ir)?|i(?:temRef|n2|s)|t(?:arget[XY]|o)|srcDoc|s(?:coped|te(?:m[hv]|p)|pan)|(?:width|size)s|prefix|typeof|itemID|s(?:t(?:roke|art)|hape|cope|rc)|t(?:arget|ype)|(?:stri|la)ng|a(?:ccept|s)|m(?:edia|a(?:sk|x)|in)|x(?:mlns)?|width|value|size|href|elementtiming|k(?:ey)?|end|low|by|i[dn]|y[12]|g[12]|x[12]|f[xy]|[yz])$/;
```
ilyabrower, line 22: this is the simplest diff as I've ever seen =)

semcore/date-picker/src/components/Calendar.jsx
msereniti: It goes good in case if person in process of selecting, after clicking first calendar unit something like "selecting date range from x"

Can't imagine other cases when `value[1]` not available

semcore/date-picker/src/components/Calendar.jsx
msereniti: It's a fail

semcore/date-picker/src/components/Calendar.jsx
msereniti: No, `ScreenReaderOnly` isn't based on Root magic and doesn't accepts DOM props

semcore/utils/src/propsForElement.ts
msereniti: Search for `abbr` keyword

msereniti: This is shit, I agree

semcore/date-picker/src/components/Calendar.jsx
ilyabrower: ok, but maybe we should change order of compoents and put select with `aria-live` in `ScreenReaderOnly`?

msereniti: Sure, that's going to be more correct

ilyabrower (approved):

PR #1178, opened 2024-03-08
[86bxr0ayr][website] add information about cenetered Spin in Select trigger

j-mnizhek:
Just a small addition to the Select trigger description, aligning the Spin in the center when the value is unknown during loading.


msereniti:
I also pushed base trigger changes that allows to implement the designed state

ilyabrower (approved):

PR #1196, opened 2024-03-13
[86bxwjwjc][utils] use React.useId in useUID when it's avaialble

msereniti:
## Motivation and Context

On one of existing project hydration errors were faced. It was caused by different `uid` generated by `useUID` hook on the client and on the server. I made it to use React.useId when it's available (we want to support older versions of React).

## How has this been tested?

I've modified code of library in node_modules of that project and ensured that hydration errors have gone.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/utils/src/uniqueID.ts
+++ b/semcore/utils/src/uniqueID.ts
@@ -19,7 +19,7 @@ const createSource = (prefix = 'ui-kit-'): ContextType => ({ value: 1, prefix })
 
 export const useUID = (prefix?: string): string => {
   const context = register.get<ContextType>('uid-context', createSource(prefix));
-  const [uid] = React.useState<number>(context.value++);
+  const [uid] = React.useId ? [React.useId()] : React.useState<number>(context.value++);
 
   useEnhancedEffect(() => {
     register.set<ContextType>('uid-context', context);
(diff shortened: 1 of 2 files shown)


semcore/utils/src/uniqueID.ts
```diff
@@ -19,13 +19,15 @@ const createSource = (prefix = 'ui-kit-'): ContextType => ({ value: 1, prefix })
 
 export const useUID = (prefix?: string): string => {
   const context = register.get<ContextType>('uid-context', createSource(prefix));
-  const [uid] = React.useState<number>(context.value++);
+  const [uid] = React.useId ? [React.useId()] : React.useState<number>(context.value++);
+  const trimmedUid =
+    String(uid).startsWith(':') && String(uid).endsWith(':') ? String(uid).slice(1, -1) : uid;
```
ilyabrower, line 24: What if `React.useId()` will add some another incorrect symbol? Could we use `CSS.escape()` on server-side?

msereniti: React useId always adds `:` to make it unusable as css selector.

> Could we use CSS.escape() on server-side?

I would prefer to keep simple slice instead

ilyabrower (approved):

PR #1220, opened 2024-03-19
[86bxtqayu][utils] fixed condition to check on which element to move focus to in a locked container

ilyabrower:
## Motivation and Context
I've changed a condition to move focus to the last element in locked container - for now we check that current active node is follow previous active node and current active node is a descendant of the previous node.

## How has this been tested?
I've tested it manually

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/utils/src/use/useFocusLock.ts
+++ b/semcore/utils/src/use/useFocusLock.ts
@@ -147,13 +147,22 @@ const useFocusLockHook = (
   const returnFocus = React.useCallback(() => {
     const trapNode = trapRef.current;
     if (trapNode && !isFocusInside(trapNode)) return;
+    const focusMastersStackCount = focusMastersStack.length;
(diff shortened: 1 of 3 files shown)


semcore/utils/src/use/useFocusLock.ts
```diff
@@ -219,6 +206,30 @@ const useFocusLockHook = (
     };
   }, [disabled, autoFocus, returnFocusTo, returnFocus]);
 
+  React.useEffect(() => {
```
msereniti, line 209: Why did you move the effect?

ilyabrower: because I need to compare values in focusMastersStack in `returnFocus` before they removed from stack

msereniti (approved):

msereniti (approved):

PR #1238, opened 2024-03-25
[86bxwjwjc][portal] removed initial mount node

msereniti:
## Motivation and Context

Commonly react frameworks [recommend](https://github.com/vercel/next.js/blob/canary/examples/with-portals/components/ClientOnlyPortal.js) to delay mount of portals with state + effect to require a state change to prevent ssr hydration errors (when we have `null` on server but portal on client).

IDK why we didn't followed this rule of thumb before and made it to return `getNodeByRef(container)` initially on server. It was causing hydration errors at least in nextjs and astro.

## How has this been tested?

Tooltip was causing hydration errors in an astro project. After this update all errors gone.

## Types of changes

- [ ] Nice improve.

## Checklist:


ilyabrower (approved):

semcore/popper/__tests__/popper.browser-test.tsx
```diff
@@ -9,17 +9,18 @@ test.describe('Popper', () => {
     await page.setContent(htmlContent);
     await page.mouse.click(1, 1);
 
+    await page.keyboard.press('Tab');
+
     for (let i = 0; i < 50; i++) {
       await page.keyboard.press('Tab');
+      await page.waitForFunction(() => {
```
msereniti, line 16: Here I've made tests stricter cause before they were passing even when component wasn't working properly

ilyabrower (approved):

PR #1278, opened 2024-04-09
[86bxmp5xu] Ellipsis with links

msereniti:
## Motivation and Context

It was found that 

```
<Ellipsis tag={Link} trim='middle' href='https://developer.semrush.com/intergalactic/'>
  https://developer.semrush.com/intergalactic/
</Ellipsis>
```

Renders as not link.

```
<Link tag={Ellipsis} trim='middle' href='https://developer.semrush.com/intergalactic/'>. https://developer.semrush.com/intergalactic/
</Link>
```

Doesn't renders at all.

The first case I fixed to make it render like a normal link. The second case I made to render as non-link – now it makes DX not so frustrating and allows developers to find a better approach of the component use. 

## How has this been tested?

Mainly manually on many cases and on a few cases with unit tests.

## Types of changes

- [ ] Nice improve.

## Checklist:


tools/testing-utils/snapshot.tsx
```diff
@@ -91,6 +92,8 @@ export const snapshot = async (
         </body>
     </html>`;
 
+  await fs.writeFile('index.html', html, 'utf8');
```
ilyabrower, line 95: looks like you needed it for debug your case...
if we will save `html` locally, let's move it to some `tmp` folder, m?

ilyabrower (approved):

tools/testing-utils/snapshot.tsx
msereniti: Yes, it was a debug statement, thank you for noticing

PR #1298, opened 2024-04-15
[86by9gyu5][select, popper] fixed role for trigger

ilyabrower:
## Motivation and Context
We've had a check about tag in Trigger and set `role=compobox` for `Select` only. I've moved this logic in `Popper` for fix `DropDown(Menu)`.

## How has this been tested?
Manyally and with our unit tests

## Types of changes

- [X] Nice improve.

## Checklist:


semcore/popper/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [5.26.4] - 2024-04-15
+
+### Fixed
+
+- `role` in Trigger, depending on the tag.
```
msereniti, line 9: I would write something like

```
### Changed

- The default trigger `role` is `combobox` if an input is used as the trigger, and button otherwise. 
```

msereniti (approved):
I've proposed probably a better changelog but final text is up to you 🙌

semcore/popper/CHANGELOG.md
ilyabrower: fixed

PR #1324, opened 2024-04-23
[website] update TabLine and TabPanel examples and a11y guides

j-mnizhek:
## Motivation and Context

We need to update:
- examples for TabLine and TabPanel, add tabpanels with content to all of them
- update documentation they have on accessibility

## How has this been tested?

Manually.

## Types of changes

- [x] Nice improve.

## Checklist:


sheila-semrush (changes requested):
I'd replace this:

> By default, the component switches content automatically when switching tabs.

With something like this:

> By default, tabs are switched automatically when selected.

Also, do we want to make showing tooltips obligatory for truncated tab names? If yes, then maybe we should remove the tip about showing the tooltip and instead implement this behaviour out of the box.

BTW, thanks for nice examples! :3

j-mnizhek:
Added small change from https://github.com/semrush/intergalactic/pull/1340 to example with disabled tab.

j-mnizhek:
Moved these changes to https://github.com/semrush/intergalactic/pull/1368

PR #1325, opened 2024-04-24
Update/breadcrumbs a11y documentation

j-mnizhek:
## Motivation and Context

We need to fix some minor issues with Breadcrumbs regarding accessibility. In this PR I've fixed a little:
- live examples
- documentation on the Design A11y and Code pages

## How has this been tested?

Manually and locally.

## Types of changes

- [x] Nice improve.

## Checklist:

### DIFF
--- a/website/docs/components/breadcrumbs/breadcrumbs-a11y.md
+++ b/website/docs/components/breadcrumbs/breadcrumbs-a11y.md
@@ -22,15 +22,19 @@ The list below describes roles and attributes that component already has.
 
 Table: Roles and attributes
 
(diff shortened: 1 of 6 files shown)


sheila-semrush (changes requested):
1. It seems that aria-label should belong to the "nav" tag.
2. We probably should make tooltip mandatory for breadcrumbs when they're truncated. If so, then <Tooltip /> should be removed from the truncation example, and the tip about showing a tooltip with the full text is also unnecessary.

sheila-semrush (approved):
Thanks!
Discussed that for now we're keeping <Tooltip /> in the recommendations and examples, but will remove it later, when corresponding changes in code will have been made.

PR #1355, opened 2024-05-15
[website] update a11y page for WidgetEmpty

j-mnizhek:
Small update to Accessibility documentation for WidgetEmpty component. 

I was thinking whether we should remove the entire 'What component has' section (as we were discussing), but it seems better to leave it because it shows what the component already has and what our users can check and test while using this component in their projects.


sheila-semrush (changes requested):
Thanks! I'd suggest a couple of things:

1. Maybe instead of merging Role and Element into one column, merge Element with Component (so `img` would be in one column with `WidgetEmpty` and `Image`), and Role with Attributes, so in the 1st row it would be like this:

> role="status", implicit aria-live="polite"... 

2. Remove "aria-hidden='true'" from the description column to reduce duplication.

j-mnizhek:
> Thanks! I'd suggest a couple of things:
> 
> 1. Maybe instead of merging Role and Element into one column, merge Element with Component (so `img` would be in one column with `WidgetEmpty` and `Image`), and Role with Attributes, so in the 1st row it would be like this:
> 
> > role="status", implicit aria-live="polite"...
> 
> 2. Remove "aria-hidden='true'" from the description column to reduce duplication.

Totally agree! Will update

PR #1358, opened 2024-05-16
[86by8j89h][docs] Add Flags a11y guide and example

sheila-semrush:
## Motivation and Context

1. Added an a11y recommendation that `aria-label` and `role` should be used when using `Flags` without the full country name.
2. Added a corresponding example right to the A11y tab — I think it's not worth complicating this guide's structure while it still doesn't have that much information.

**IMPORTANT.** This PR already assumes that `Flags` doesn't have an `aria-label`, so it should be merged only when the `aria-label` is actually removed from the component.

## How has this been tested?

Manually.

## Screenshots:

image

## Types of changes

- [x] Nice improve.

## Checklist:


website/docs/components/flags/examples/aria-label.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const Demo = () => (
+  <>
+    {countries.map((country) => (
+      <Flags
+        key={country}
+        name={country as keyof typeof iso2Name}
+        role='image'
+        aria-label={iso2Name[country]}
+        mr={1}
```
j-mnizhek, line 16: I would suggest to add mb={3} here just to visually separate first row with the following column a little bit. Something like this:
image

j-mnizhek: Great example btw!

j-mnizhek (approved):
Added small suggestion. Everything other is great in my opinion ✨

website/docs/components/flags/examples/aria-label.tsx
sheila-semrush: Added the margin, thanks!

msereniti (approved):

PR #1462, opened 2024-06-24
[86bydf6qv][feed-back] Fix feedback form a11y

ilyabrower:
Some fixes in Feedback Form for a11y


msereniti (approved):

sheila-semrush (changes requested):
Discussed to reduce tooltips width to 228px

sheila-semrush (approved):
thanks! also updated the max-width in Tooltip design guide

PR #1471, opened 2024-06-27
[notice] removed critical notification label from notice with warning theme

msereniti:


sheila-semrush:
@msereniti can we do the same for NoticeGlobal in this PR?

ilyabrower (approved):

PR #1480, opened 2024-06-28
[website] update InputPhone docs

j-mnizhek:
## Motivation and Context

- Updated InputPhone examples a little bit.
- Fixed accessibility documentation for InputPhone a little bit.

## How has this been tested?

Locally.

## Types of changes

- [x] Nice improve.

## Checklist:


website/docs/components/input-phone/examples/known_country_but_the_number_format_is_unknown.tsx
```diff
@@ -18,12 +18,7 @@ const Demo = () => {
         </Input.Addon>
         <Input.Value value={value} onChange={(v) => setValue(v)} id='phone-number' />
         {Number.parseInt(value, 10) > 2 && (
-          <Input.Addon
-            tag={CloseM}
-            interactive
-            aria-label='Clear field'
-            onClick={() => setValue('+1')}
-          />
+          <Input.Addon tag={CloseM} interactive aria-label='Clear' onClick={() => setValue('+1')} />
```
sheila-semrush, line 21: shouldn't we update aria-labels in the other 2 examples as well?

website/docs/components/input-phone/input-phone-a11y.md
```diff
@@ -4,38 +4,10 @@ a11y: AA
 tabs: Design('input-phone'), A11y('input-phone-a11y'), Example('input-phone-code')
 ---
 
-## What component has
+This component consists of Input and Select elements, so feel free to check their detailed documentation regarding accessibility:
```
sheila-semrush, line 7: should the InputMask be mentioned here as well? or we'll do it after we fix its a11y?

sheila-semrush, line 7: wouldn't it be better to call "Input and Select" components instead of elements?
and also, is it correct to call InputPhone a component? It looks more like a pattern. We might have reasons not to move it away from other Input* components, but maybe it'll be safer to avoid the word "component" when mentioning it?

sheila-semrush (changes requested):
Suggested a couple of changes

website/docs/components/input-phone/input-phone-a11y.md
j-mnizhek: > wouldn't it be better to call "Input and Select" components instead of elements?
OMG, yes, of course, it's a typo. I'll fix this.

> and also, is it correct to call InputPhone a component? It looks more like a pattern. We might have reasons not to move it away from other Input* components, but maybe it'll be safer to avoid the word "component" when mentioning it?
Good point! 👍 Agree on this so much.

j-mnizhek: This is a good point! I wasn't sure should we add since not all the examples use it. But looks like it could be a good addition.

sheila-semrush (approved):
Looks great, thanks!

ilyabrower:
See #1502

PR #1503, opened 2024-07-09
[86byz6xq0][tooltip] added properties for paint arrow

ilyabrower:
## Motivation and Context
I've added two props for Tooltip.Popper to determine the color of the arrow.

## How has this been tested?
Manually and I've added screenshot test

## Types of changes

- [X] Nice improve.

## Checklist:


j-mnizhek (changes requested):
- Asked two questions in the inner chat.
- Updated example a little bit ✨ Waiting for possibility to pass tokens to the new properties. Will approve as soon as it will work 😎

j-mnizhek (approved):
Updated example for colored Tooltip. Everything other looks good!

semcore/tooltip/src/Tooltip.jsx
```diff
@@ -171,9 +173,15 @@ function TooltipPopper(props) {
           use:theme={resolveColor(theme)}
           use:aria-live={undefined}
           ref={popperRef}
+          __excludeProps={['arrowBgColor', 'arrowShadowColor']}
```
msereniti, line 176: It should not be applied anyway due to propsForElement filter

semcore/tooltip/src/index.d.ts
```diff
@@ -7,6 +7,11 @@ import Popper, {
   eventInteraction,
 } from '@semcore/popper';
 
+type ArrowCustom = {
```
msereniti, line 10: I guess we should export it

semcore/tooltip/src/style/tooltip.shadow.css
```diff
@@ -72,6 +72,14 @@ SArrow[theme='warning'] {
   }
 }
 
+SArrow[bgColor][shadowColor] {
```
msereniti, line 75: I'm not sure it will be good to set them only when both are provided while it's two optional properties. Maybe we should provide it in one prop as object?

semcore/tooltip/src/Tooltip.jsx
ilyabrower: removed

semcore/tooltip/src/index.d.ts
ilyabrower: added

semcore/tooltip/src/style/tooltip.shadow.css
ilyabrower: `shadowColor` has a fallback to `bgColor`, so, here it will always be defined if `bgColor` is defined

msereniti (approved):

PR #1576, opened 2024-08-09
[86bzwpc3r][illustration] Added MechanismInside illustration

sheila-semrush:
## Motivation and Context

Added a new illustration due to a request.

Added it in the end of the "States" category:

image

##

Tested with different colors and sizes:

image

##

Also sending a branch with this illustration to review in Figma.

## Types of changes

- [ ] Nice improve.

## Checklist:


website/docs/style/illustration/illustrations-list.js
```diff
@@ -32,6 +32,10 @@ const illustartionsList = {
       name: 'UnderConstruction',
       group: 'States',
     },
+    {
+      name: 'MechanismInside',
```
j-mnizhek, line 36: What do you think if we call it something like Settings, or Broken, or maybe with the name of the state it helps to visualise? Why I'm thinking about it: all other illustrations in this group have names of the states so it's clear for the users when they can use them.
image

j-mnizhek (approved):
Left comment only about the illustration name. Everything other seems good! ✨

website/docs/style/illustration/illustrations-list.js
sheila-semrush: sounds reasonable, I think we can call it "Configuration" (the team that wants to use it now uses it as "please configure something"

PR #1590, opened 2024-08-13
[86bzwmvur][button] changed hints in buttons with addons only

ilyabrower:
## Motivation and Context
I've changed Hints in buttons with only addons:
- the trigger for now is the button instead of addon
- fixed offset for popper
- added snapshot tests and VO
- fixed example in documentation

## How has this been tested?
Manually and with tests

BEFORE:
Screenshot 2024-08-13 at 16 03 47
AFTER:
Screenshot 2024-08-13 at 16 03 35

## Types of changes

- [X] Nice improve.

## Checklist:


sheila-semrush (commented):
Ok, so the offset seems correct, everything looks fine. 
But I just realized that this aria-label magic doesn't show the Hint on focus (it should).
(This should've been in the task initially, I just haven't noticed before.)

image

ilyabrower:
> Ok, so the offset seems correct, everything looks fine. But I just realized that this aria-label magic doesn't show the Hint on focus (it should). (This should've been in the task initially, I just haven't noticed before.)
> 
> image

Fixed it!

sheila-semrush (approved):

msereniti (approved):

PR #1596, opened 2024-08-14
[86bzt3f3m][wizard] enhanced a11y

msereniti:
## Motivation and Context

This PR fixes following a11y related things:
1. Instead of relying on aria-label, now wizard will use aria-labelledby to refer to header
2. Now menu uses tabs+tabpanel pattern.
3. I added built in back and next buttons with aria-labels
4. I have content autofocus on steps change

## How has this been tested?

Manually.

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/wizard/src/index.d.ts
```diff
@@ -62,6 +64,17 @@ export type WizardStepperProps<T extends WizardStep = WizardStep> = BoxProps & {
   disabled?: boolean;
 };
 
+export type WizardStepBackProps<T extends WizardStep = WizardStep> = ButtonProps & {
+  onActive?:
```
ilyabrower, line 68: what do you think about simple onClick with two arguments - event and step ?

semcore/wizard/src/Wizard.jsx
```diff
@@ -142,20 +186,72 @@ function Stepper(props) {
...
+  const { Children, styles, getI18nText, step } = props;
+  const handleClick = React.useCallback(() => {
+    props.onActive?.(props.step - 1);
+  }, [props.step]);
+  return sstyled(styles)(
+    <SStepBack
+      render={Button}
+      use='tertiary'
+      onClick={handleClick}
+      aria-label={getI18nText('backButton', { buttonName: `#${step + 1}` })}
```
sheila-semrush, line 214: aria-label should contain the button text instead of the step number
e.g. "Back to Location", "Go to Schedule"
image

sheila-semrush (changes requested):
1. I don't see this one, I still see "Modal window" as the wizard's aria-label

> "Instead of relying on aria-label, now wizard will use aria-labelledby to refer to header"

2. The heading in the sidebar is set using Sidebar's title property, which is interpreted as an html title and appears on hover throughout the whole sidebar. Screen reader also announces the whole sidebar using this title ("Site Audit Settings, group"). I think it's incorrect, maybe we should supress passing the "title" attribute to the rendered html element?

image

sheila-semrush (changes requested):
All my inline comments were fixed, thanks! I've marked them resolved.
The above comment is still relevant (about modal's aria-label and sidebar title).
And also now there's something wrong with the StepNext's text, it's not rendered at all:
image

semcore/wizard/src/Wizard.jsx
```diff
@@ -43,42 +74,48 @@ class WizardRoot extends Component {
...
-  const { Children, styles, title } = props;
+  const { Children, styles, title, uid } = props;
   const SSidebar = Root;
   const SSidebarHeader = 'h2';
+  const SSidebarMenu = 'div';
   return sstyled(styles)(
-    <SSidebar render={Box} role='menu'>
-      {title && <SSidebarHeader>{title}</SSidebarHeader>}
-      <Children />
+    <SSidebar render={Box}>
```
sheila-semrush, line 114: Sidedar has an HTML title that equals to the text of the Sidebar heading. Because of that there's always a system tooltip when my mouse is over the Sidebar. And the Screen reader announces the Sidebar using this title, which is incorrect, because this title describes the whole modal, not the Sidebar. 

Can we suppress passing the "title" attribute to the resulting div element?

image image

semcore/wizard/src/Wizard.jsx
```diff
@@ -124,7 +171,8 @@ function Stepper(props) {
   return sstyled(styles)(
     <SStepper
       render={Box}
-      role='menuitem'
+      role='tab'
+      aria-controls={active ? `${uid}-content-${step}` : undefined}
```
sheila-semrush, line 175: sorry, I just realized we should've replaced `aria-current` with `aria-selected`! `aria-selected` should be used in tablists

semcore/wizard/src/Wizard.jsx
```diff
@@ -142,20 +190,70 @@ function Stepper(props) {
...
-  const { Children, styles } = props;
+  const { Children, children: hasChildren, styles, uid, step } = props;
   const SContent = Root;
   return sstyled(styles)(
-    <SContent render={Box}>
-      <Children />
+    <SContent
+      render={Box}
+      role='tabpanel'
+      aria-labelledby={`${uid}-step-${step}`}
```
sheila-semrush, line 199: it refers to the wrong element (the tabpanel)
it should refer to the current Stepper (tab)

sheila-semrush (commented):
@j-mnizhek can you look at the docs & examples please?
and we're waiting for the answer regarding the aria-orientation / tab navigation

semcore/wizard/src/Wizard.jsx
```diff
@@ -124,9 +171,11 @@ function Stepper(props) {
   return sstyled(styles)(
     <SStepper
       render={Box}
-      role='menuitem'
+      role='tab'
```
sheila-semrush, line 174: @msereniti after discussion with Glafira, it seems we have to add navigation with Arrows.
In this case it should be Up/Down Arrow.
But IMO the tabs shouldn't activate on selection, i.e. user has to choose with an Arrow and then confirm with Enter/Space (like in Tabline/Tabpanel with behavior=manual).

semcore/wizard/src/Wizard.jsx
```diff
@@ -164,8 +179,14 @@ function Stepper(props) {
         e.preventDefault();
         onActive(step, e);
       }
+      if (e.key === 'ArrowUp') {
```
sheila-semrush, line 182: arrow navigation works well!
but we should remove navigation through the tablist with the Tab key, Tab should move focus out of the tablist

PR #1601, opened 2024-08-16
[chore] updated some lost translations

ilyabrower:


msereniti (changes requested):
Need changelog not to every cahnge :(
Please, exclude wizzard, tag and input tag changes as it's upcomming from other PRs

msereniti (approved):

PR #1610, opened 2024-08-21
[86c00j9tv][tooltip] in some rare edge cases, tooltip were not appearing, also tooltip with ellipsis in breadcrumbs example

msereniti:
## Motivation and Context

It was found that tooltips wasn't showing in breadcrumbs example. I found a weird css code that relies on `data-popper-reference-hidden`. I can't understand the point of this attr – will se if tests are not broken.

Also I replaced tooltips with ellipsis, to use more correct aria attributes and show popper only when text is cut.

## How has this been tested?

Manually.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/website/docs/components/breadcrumbs/examples/usage_example.tsx
+++ b/website/docs/components/breadcrumbs/examples/usage_example.tsx
@@ -11,7 +11,7 @@ const Demo = () => (
       <Ellipsis.Popper />
     </Ellipsis>
     <Ellipsis>
-      <Ellipsis.Content tag={Breadcrumbs.Item} href='#' role={'link'}>
+      <Ellipsis.Content tag={Breadcrumbs.Item} active={false} href='#' role={'link'}>
         This title is longer than a giraffe's neck, I bet it's been doing neck workouts!
       </Ellipsis.Content>
       <Ellipsis.Popper />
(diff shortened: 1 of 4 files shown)


sheila-semrush (commented):
Doesn't work for me :( both in chrome and firefox
Tooltip doesn't appear, the first item is cropped without the "...", and the links are not underlined on hover
[video](https://github.com/user-attachments/assets/037e006a-15c2-4b36-b1d7-4d20b878d986)

sheila-semrush (changes requested):
(see comment above)

website/docs/components/breadcrumbs/examples/usage_example.tsx
```diff
@@ -1,21 +1,21 @@
...
-      This title is longer than a giraffe's neck, I bet it's been doing neck workouts!
-    </Tooltip>
+    <Ellipsis>
+      <Ellipsis.Content tag={Breadcrumbs.Item} active={false} href='#' role={'link'}>
+        Ellipsis
+      </Ellipsis.Content>
+      <Ellipsis.Popper />
+    </Ellipsis>
+    <Ellipsis>
+      <Ellipsis.Content tag={Breadcrumbs.Item} href='#' role={'link'}>
```
sheila-semrush, line 14:
```suggestion
      <Ellipsis.Content tag={Breadcrumbs.Item} active={false} href='#' role={'link'}>
```
I guess it needs to be added as in the 1st item, because otherwise the link behaves weirdly (not underlined on hover)

sheila-semrush (approved):
It works, hurray!
Only one little change needs to be made in the example, I left an inline suggestion.

PR #1626, opened 2024-08-28
[86c00cvw1][date-picker] removed input-mask

msereniti:
Will extend description later

### DIFF
--- a/semcore/date-picker/src/components/InputTrigger.jsx
+++ b/semcore/date-picker/src/components/InputTrigger.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import BaseTrigger from '@semcore/base-trigger';
-import InputMask from '@semcore/input-mask';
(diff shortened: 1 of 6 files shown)


semcore/date-picker/src/components/InputTrigger.jsx
```diff
@@ -843,29 +656,160 @@ const MaskedInput = ({
     setAppliedWidth(width);
   }, [width, focused]);
 
+  const maskInstanceRef = React.useRef(null);
+  React.useEffect(() => {
+    if (!ref.current) return;
+    if (maskInstanceRef.current) return;
```
ilyabrower, line 662: `!maskInstanceRef.current` ?

msereniti: No, it's to prevent unexpected reinitialization

PR #1632, opened 2024-09-04
[86bzw6xkg][date-picker] calendar as accessible grid

msereniti:
## Motivation and Context

This PR makes our calendars of all date-p family to be represneted as grid instead of long list of buttons.

I fixed a11y of custom day example as well.

## How has this been tested?

Manually with VO. Needs testing with NDVA

## Types of changes

- [x] Nice improve.

## Checklist:


semcore/date-picker/src/components/Calendar.jsx
```diff
@@ -329,13 +343,15 @@ class CalendarDaysRoot extends CalendarAbstract {
     const description = this.describeValue();
 
     return sstyled(styles)(
-      <SCalendar render={Box}>
-        <CalendarWeekDays locale={locale} />
-        <SGridDays onMouseLeave={this.handleMouseLeave}>
+      <SCalendar render={Box} {...this.getRootProps()}>
+        <CalendarWeekDays locale={locale} role='row' />
```
ilyabrower, line 347: why do we need `row` here? Should we `walk` through the header of Calendar?

ilyabrower (approved):

semcore/date-picker/src/components/Calendar.jsx
msereniti: It's required to put row between grid and cell

j-mnizhek:
Have a list of questions and comments, but asked about the PR purpose in the DM first, because maybe some of them are irrelevant.

Here are some first comments and issues I've found:

1. In the DatePicker example, the combined aria-label is read incorrectly: From the label (for= attribute): "Date range picker Month range picker". The "Month range picker" needs to be removed. There might be an issue with the uniqueness of IDs on the page.

2. In the examples with presets, the focus order in the dialog is different. In simple examples like DateRangePicker, the focus starts with the preset buttons, while in the pickers with period comparison, it starts with the month switching buttons. Can we use a consistent focus order to avoid confusing users? For example, starting with the month switching buttons?

3. Can we use `columnheader` for the weekday cells? They should then be read together with the days.

4. Can we set `aria-current="date"` on the cell for the selected date if it’s the current date?

5. We should add a `visually-hidden` `span` with the full name of the week for the weekdays.

6. In the WeekPicker, for some reasons I can’t navigate through the calendar using the keyboard 😔

msereniti:
> In the DatePicker example, the combined aria-label is read incorrectly: From the label (for= attribute): "Date range picker Month range picker". The "Month range picker" needs to be removed. There might be an issue with the uniqueness of IDs on the page.

Fixed the id

> In the examples with presets, the focus order in the dialog is different. In simple examples like DateRangePicker, the focus starts with the preset buttons, while in the pickers with period comparison, it starts with the month switching buttons. Can we use a consistent focus order to avoid confusing users? For example, starting with the month switching buttons?

Fixed

> Can we use columnheader for the weekday cells? They should then be read together with the days.

Changed

> Can we set aria-current="date" on the cell for the selected date if it’s the current date?

Changed

> We should add a visually-hidden span with the full name of the week for the weekdays.

Added aria-label instaed of  visually-hidden span. I guess it's a better option because it will not affect select-&-copy content but will replace the shorten version. Also I've removed attr attribute because we were using it in a reversed way – it should contain shorted version but we were putting the longer one.

> In the WeekPicker, for some reasons I can’t navigate through the calendar using the keyboard 😔

Updated example, it was very 💩

j-mnizhek:
Back here! Found some issues:

1. Cannot submit the choosen date from the calendar grid by keyboard. I'm pressing Enter on the screencast below, while I'm on some date, but nothing happens.
Sep-18-2024 11-47-02

2. MonthRangePicker reads months as Month 1. Shouldn't just the month be read in this case? It's a bit confusing at the moment.
image

3. Btw, can we change aria-labels for Prev and Next buttons for different DatePicker types? What I'm talking about:
- For all of them they now are read as **Previous period** and **Next period**, but it looks like the following labels will be more convenient:
--- for DatePicker, DateRangePeaker, Week picker: **Previous month** and **Next month**
--- for MonthPeaker and MonthRangePeaker: **Previous year** and **Next year**

j-mnizhek (approved):
Looks like we have all the necessary changes and fixes 😎

website/docs/components/date-picker/date-a11y.md
```diff
@@ -7,31 +7,86 @@ tabs: Design('date-picker'), A11y('date-a11y'), API('date-api'), Example('date-c
...
+| <nobr>`DatePicker.Trigger` > `input`</nobr> | `aria-haspopup="dialog"` and inherits `aria-controls="IDREF"` from its parent. | `aria-haspopup` indicates that the element triggers a dialog.                                                  |
+|                                             | `aria-expanded="false/true"`                                                   | Indicates whether the dialog is open or not.                                                                   |
+|                                             | `role="combobox"`                                                              | Identifies the element as a combobox—a text input combined with a list of options.                             |
+|                                             | `aria-describedby="IDREF"`                                                     | Provides the input with an accessible description by linking to a hint that explains the required data format. |
```
sheila-semrush, line 41: don't get this thing with the hint, in the basic example it refers to an empty span. Is it correct? How is it supposed to work?

sheila-semrush:
@j-mnizhek I updated the Keyboard support table and added a couple other small fixes. You can review my last commit and edit over it if you want. Everything looks fine, only have one question about the hint (another comment).

website/docs/components/date-picker/date-a11y.md
j-mnizhek: Yeah, for sure, it's strange. The date format "MM/DD/YYYY" should be moved from aria-label to that span

image

j-mnizhek: Guys, @ilyabrower @msereniti can we move date format to the span that should contain the date format?

j-mnizhek: Michael fixed this ✅

PR #1644, opened 2024-09-12
[86bzhabjx][select] Refactored select with new dropdown's and some a11y improvements

ilyabrower:
## Motivation and Context
I've added some `AbstractDropdown` class for extend Select and DropdownMenu from them.
I've added back example with nested dropdown menu and focusable elements.

## How has this been tested?
Manually and with existing tests

## Types of changes

- [X] Nice improve.

## Checklist:


sheila-semrush (commented):
Found an issue with focus, when there's a Notice in the Select: if there are interactive elements in the Notice, I can focus them by pressing Tab, but then I get stuck. 

Discussed that it would be nice if after focusing the control in the Notice, I could press Tab again or Shift-Tab and return to the same state as before (the Trigger is focused, and I can select options with Arrows).

Also have a couple of questions for Hlafira, will ask her and get back with answers.

ilyabrower:
> Found an issue with focus, when there's a Notice in the Select: if there are interactive elements in the Notice, I can focus them by pressing Tab, but then I get stuck.

I fixed it the way it was - we'll get the focus back on the trigger if we press the arrows.

semcore/dropdown/src/AbstractDropdown.tsx
```diff
@@ -0,0 +1,347 @@
...
+    const triggerId = this.triggerRef.current?.id;
+
+    return {
+      size,
+      index: this.asProps.highlightedIndex,
+      tabIndex: -1,
+      ref: this.menuRef,
+      id: `igc-${uid}-list`,
+      role: this.role,
+      'aria-labelledby': triggerId,
```
sheila-semrush, line 116: listbox should have the same label as the Select.Trigger (ideally)
but if that's too complicated, we can make it a static text string (like before)

sheila-semrush:
Found a serious issue: in the example with Search option labels are not announced properly, "Section selected" is announced on each list item by NVDA.

Btw, the following is not a problem by itself, but just a weird thing that's probably related: when I focus the whole listbox (by pressing Tab), NVDA announces it as "Section", and I don't understand why, because there shouldn't be any sections.

msereniti (approved):

semcore/dropdown/src/AbstractDropdown.tsx
ilyabrower: I made ideally option, but check it please

ilyabrower:
> 1. Found a serious issue: in the example with Search option labels are not announced properly, "Section selected" is announced on each list item by NVDA.
> 2. Btw, the following is not a problem by itself, but just a weird thing that's probably related: when I focus the whole listbox (by pressing Tab), NVDA announces it as "Section", and I don't understand why, because there shouldn't be any sections.
> 
> Other issues with keyboard navigation:
> 
> 3. In the example with Notice and Link, I can enter the list only after I focus the link with Tab, and then press Shift-Tab. And if I remove the link, I can't enter the list and select options at all. What's even weirder, I though it wasn't like that before the last change, but after rolling it back I have the same behavior (I did pnpm install and restarted the server).
> 4. Now I have to use Tab or Shift-Tab to move between search and listbox, but as soon as I press Tab, I get stuck. The listbox is selected entirely, and the Arrow keys don't work. I can go back to Search with Shift-Tab, but to go from Search to the listbox I have to use strictly arrows, to not get stuck.

I've fixed a little aria* attributes on trigger and for now, NVDA works well. Recheck it please

msereniti (approved):

semcore/select/src/Select.jsx
```diff
@@ -81,20 +105,29 @@ class RootSelect extends Component {
...
-      'aria-haspopup': 'listbox',
-      'aria-expanded': visible ? 'true' : 'false',
+      ...super.getTriggerProps(),
+      onKeyDown: callAllEventHandlers(
+        this.handlePreventCommonKeyDown.bind(this),
+        this.handleOpenKeyDown.bind(this),
+        this.handleArrowKeyDown.bind(this),
+      ),
+      'aria-controls': visible ? ariaControls : undefined,
+      'aria-haspopup': isMenu ? 'listbox' : 'dialog',
```
sheila-semrush, line 126: I suggest making it always `aria-haspopup="listbox"`. 

We don't have examples with focusable elements other than the Search field, so in our examples users don't have to use Tab in Select.Popper. We'll add it as a recommendation, to change the value from "listbox" to "dialog" whenever they have something focusable other than the search field.

website/docs/components/select/examples/options_filtering.tsx
sheila-semrush, line 1: is it reasonable to leave all ARIA attributes for the InputSearch for developers to add each time by themselves (lines 37-42)? Can we set them out of the box?

semcore/select/src/Select.jsx
sheila-semrush: especially since we now always have aria-activedescendant, regardles the popper structure, if I understand correctly

sheila-semrush:
http://localhost:5173/intergalactic/components/dropdown-menu/dropdown-menu-api crashes for me
image

website/docs/components/select/examples/options_filtering.tsx
sheila-semrush: (maybe we need something like a "searchable" prop for Select, instead of having to insert the InputSearch component each time)

semcore/select/src/InputSearch.jsx
```diff
@@ -102,12 +101,8 @@ const SearchValue = (props) => {
 };
 const SearchClear = (props) => {
```
sheila-semrush, line 102: The Hint is squished horizontally
image

semcore/dropdown/src/AbstractDropdown.tsx
sheila-semrush: works like magic!!!

sheila-semrush:
> > 1. Found a serious issue: in the example with Search option labels are not announced properly, "Section selected" is announced on each list item by NVDA.
> 
> I've fixed a little aria* attributes on trigger and for now, NVDA works well. Recheck it please

Thanks! 2-3 seem to be fixed, but 1 still reproduces in Firefox, unfortunately (I've run install & build). In Chrome it works fine.
image

website/docs/components/select/examples/loading_state.tsx
```diff
@@ -20,7 +20,7 @@ const Demo = () => (
       </Text>
       <div>
         <Select mt={2} mr='auto' id='loading-select-no-chevron'>
-          <Select.Trigger chevron={false} placeholder={<Spin size='xs' mx={4} />} />
+          <Select.Trigger chevron={false} placeholder={<Spin size='xs' mx={4} />} tabIndex={-1} />
```
sheila-semrush, line 23:
```suggestion
          <Select.Trigger chevron={false} placeholder={<Spin size='xs' mx={4} />} disabled />
```
I think it should just be disabled.

website/docs/components/select/examples/loading_state.tsx
sheila-semrush, line 1: I think "loading" prop should make the Select disabled. It will be announced more correctly by SR then.

website/docs/components/select/examples/options_filtering.tsx
ilyabrower: > (maybe we need something like a "searchable" prop for Select, instead of having to insert the InputSearch component each time)

ilyabrower: And as I see, now, we shouldn't pass all these attributes

ilyabrower:
> http://localhost:5173/intergalactic/components/dropdown-menu/dropdown-menu-api crashes for me
> 
> image

fixed this

website/docs/components/select/examples/loading_state.tsx
ilyabrower: Maybe, but there may be a case when we load something in open select. And if we set it to disabled - it will close the popper. So, I think we should collect some cases and then decide.

semcore/select/src/Select.jsx
ilyabrower: That's not how it works. 
`dialog` is a default role of popper, but in `Select.Menu` we are `cancelling` it. 
So, we could cancel also role dialog in popper in `Select.Popper`, but I'm not sure that this is a right way.

semcore/select/src/InputSearch.jsx
ilyabrower: Let's come back to this later (it's not part of Select, it's something like popper/tooltips). And it's not all trivial.

ilyabrower:
> 2\. Basic example. When Select is open and an option is highlighted, press Tab. Trigger gets focus and is announced. Pressing Tab again doesn't do anything, but I can press an Arrow and return to choosing options from the list. It's weird because navigation with Tab seems to be half-disabled - I can Tab once, but then I have to get out with Arrows.

What is your expectations?

sheila-semrush:
> > 1. Basic example. Press Tab or Shift-Tab to focus a closed Select, press Enter. Select opens, but no options are highlighted. After closing and reopening, the 1st option is highlighted. Reproduces everywhere, even in Chrome on Mac.
> 
> fixed

seems fixed with Tab, but still reproduces with Shift Tab :)

sheila-semrush:
> > 2. Basic example. When Select is open and an option is highlighted, press Tab. Trigger gets focus and is announced. Pressing Tab again doesn't do anything, but I can press an Arrow and return to choosing options from the list. It's weird because navigation with Tab seems to be half-disabled - I can Tab once, but then I have to get out with Arrows.
> 
> What is your expectations?

either Tab not doing anything, or working both ways (I can Tab from the list to the trigger and back)

sheila-semrush:
Noticed a new issue:

NVDA doesn't announce the selected/not selected status of options. Both Chrome & Firefox. To be fair, it doesn't do that in the ARIA APG example either. But it seems critical in the case with Multiselect at least. What's interesting, it works fine on prod: adds "selected" for selected and nothing for not selected (I guess this is fine since it works similarly with tabs).

ilyabrower:
> > > 1. Basic example. Press Tab or Shift-Tab to focus a closed Select, press Enter. Select opens, but no options are highlighted. After closing and reopening, the 1st option is highlighted. Reproduces everywhere, even in Chrome on Mac.
> > 
> > 
> > fixed
> 
> seems fixed with Tab, but still reproduces with Shift Tab :)

fixed again =)

semcore/select/src/Select.jsx
sheila-semrush, line 1: Bizarrely, there's now a bug that reproduces only in Chrome. When opening Select with some option already selected, the list isn't scrolled to the selected option. On the screen an option in the end of the list is selected. Works fine in Firefox.
image

sheila-semrush, line 1: I still (or again) can get stuck in the list in this position (Firefox):

image

sheila-semrush, line 1: When I open Select with search, the 1st item is initially highlighted, but I don't hear it. You can see in the NVDA log that it's supposed to be announced, but right at the same time the Search field is focused, and it interrupts the highlighted option announcement. 

This has 2 side effects:
- if I press Enter, "Apple" becomes selected, but it's unexpected because I didn't hear it being highlighted
- if I press Down Arrow, "Banana" will be highlighted, which is perceived as skipping the first option due to the same reason

image

If nothing is found, I can know it only if I switch to the "browse mode" and go read the text below the search field. Maybe we could add the "nothing found" text as `aria-describedby` to the search field? 

image

I think we should always have the `aria-describedby` when something is entered in Search. When something is found, it should refer to the SR-only text like "X results found".

website/docs/components/select/examples/options_filtering.tsx
```diff
@@ -23,38 +26,17 @@ const Demo = () => {
...
-            </>
-          )}
+          <InputSearch value={filter} onChange={setFilter} placeholder='Search' />
+          <Select.List hMax={'224px'}>
+            {options.map(({ value, label }) => (
+              <Select.Option value={value} key={value}>
+                {label}
+              </Select.Option>
+            ))}
+            {!options.length && <Select.OptionHint key='Nothing'>Nothing found</Select.OptionHint>}
```
sheila-semrush, line 31: is it ok that we use deprecated `OptionHint` in the example?

website/docs/components/select/examples/options_filtering.tsx
```diff
@@ -1,64 +1,69 @@
...
-                  <Select.OptionHint key='Nothing'>Nothing found</Select.OptionHint>
-                )}
-              </Select.List>
-            </>
-          )}
+          <InputSearch
+            value={filter}
+            onChange={setFilter}
+            placeholder='Search'
+            aria-describedby={filter && 'search-result'}
```
sheila-semrush, line 29: @ilyabrower I initially wrote a comment about announcing what's found with `aria-describedby`, but then decided to just add it myself. Can you review pls? I also wasn't sure what's the best way to pluralize "results", so edits or advice are welcome.

website/docs/components/select/examples/options_filtering.tsx
```diff
@@ -1,64 +1,69 @@
...
+          <Select.List hMax={'224px'}>
+            {options.map(({ value, label }) => (
+              <Select.Option value={value} key={value}>
+                {label}
+              </Select.Option>
+            ))}
+            {options.length ? (
+              <ScreenReaderOnly id='search-result'>{options.length} results found</ScreenReaderOnly>
+            ) : (
+              <Select.OptionHint id='search-result' key='Nothing'>
```
sheila-semrush, line 40: is it ok that we use the deprecated `OptionHint`?

website/docs/components/select/examples/options_filtering.tsx
ilyabrower: I rewrote it a little

semcore/select/src/Select.jsx
ilyabrower: I could fix it only if I'll set `tabIndex = -1` to this div, but after that we will have an error with axe `Element has children which are not allowed: div[tabindex]`
So, let's stay it as is, looks like FF bug

ilyabrower: I added setTimeout for input focusing, but you still can't hear it, but it will be in the Speach viewer

ilyabrower: fixed

ilyabrower: I added logic - if we have Selectrt.InputSearch - we don't select the first element by default. If we already have a selected option - we will select it (but won't hear it, since the focus is on the search input)

sheila-semrush: good news - the example with search works as described
bad news - it seems examples without search broke
image

sheila-semrush: works perfectly in examples with search!
will need to check in examples without search after the last bug is fixed

sheila-semrush: in lists without search now the 1st item is highlighted when I open them (with Enter) for the 1st time, and nothing is highlighted when I do it again. Actually, it's ok that nothing is highlighted if I open with Enter. But the weird thing is that it works differently the 1st and the following times.

The 1st/last option is still highlighted even in the Select with search, if I open it with Down/Up Arrow.

ilyabrower: fixed both

sheila-semrush:
@j-mnizhek take a look at the docs please. I'm still planning to update the Design tab, for now I've just updated the composition illustration.

j-mnizhek (approved):
Phew. Just finished this big journey with Select exploration.
- Made some small fixes according to vale recommendations
- Updated a couple of outdated illustrations
- Updated content in some examples for clarity

Found only problem with the keyboard interaction and scrolling to the selected item. Some others need separate tasks, because they are not in the scope of this task.

website/docs/components/select/select-a11y.md
```diff
@@ -51,13 +51,13 @@ The following list will help you to keep in mind the necessary roles and attribu
 
 Table: Roles and attributes
 
-| Component            | Attribute                       | Usage                                                                                         |
-| -------------------- | ------------------------------- | --------------------------------------------------------------------------------------------- |
-| `Select.InputSearch` | `aria-describedby="IDREF"`      | Define an accessible description announcing the number of search results. Refer to the [options filtering example](./select-code.md#options-filtering). |
+| Component            | Attribute                  | Usage                                                                                                                                                    |
+| -------------------- | -------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `Select.InputSearch` | `aria-describedby="IDREF"` | Defines an accessible description announcing the number of search results. Refer to the [options filtering example](./select-code.md#options-filtering). |
```
sheila-semrush, line 56: @j-mnizhek it was an intentional imperative, because they have to do that themselves

j-mnizhek: It’s a little bit confusing that the documentation pattern changes. In the table above, we describe what a specific role or attribute does, while in this table, we provide a recommendation.

sheila-semrush: If we change it from "define" to "defines", then the whole sentence should be rewritten, because obviously aria-describedby per se doesn't have anything to do with search results. I wanted to write a specific recommendation on how to use the property in this specific case. 

Then it should be something like "Defines an accessible description for the element. Use it to announce the number of search results." What do you think about this version?

j-mnizhek: Got it! Yeah! I think this version will be better 👍

j-mnizhek: Got it! Yeah! I think this would suit better 👍
(it's strange I've sent this comment 2 hours ago and now I see it disappeared somehow)

sheila-semrush: this PR is just too big and cursed, it happened to me as well

PR #1664, opened 2024-09-24
[86c0czv0x][d3-chart] d3 chart a11y

ilyabrower:
## Motivation and Context
I've added some roles in charts and aria-labels in examples

## How has this been tested?
Manually

## Types of changes

- [X] Nice improve.

## Checklist:

### DIFF
--- a/semcore/d3-chart/src/a11y/PlotA11yModule.tsx
+++ b/semcore/d3-chart/src/a11y/PlotA11yModule.tsx
@@ -18,7 +18,7 @@ export type A11yViewProps = {
   locale: NavigatorLanguage['language'];
   config: PartialDataSummarizationConfig;
 
-  plotRef: React.RefObject<Element>;
+  plotRef: React.RefObject<HTMLElement>;
 };
 
 export const PlotA11yModule: React.FC<A11yViewProps> = (props) => {
--- a/semcore/d3-chart/src/a11y/PlotA11yView.tsx
+++ b/semcore/d3-chart/src/a11y/PlotA11yView.tsx
@@ -52,7 +52,11 @@ export const PlotA11yView: React.FC<A11yViewProps> = ({
     }, 0);
   }, [providedData, hints, config, locale, translations, localizedMessages]);
 
-  const handleSkip = React.useCallback(() => {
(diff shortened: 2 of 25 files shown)


msereniti (approved):

semcore/d3-chart/src/translations/en.json
```diff
@@ -1,3 +1,4 @@
 {
-  "trend": "Trend"
+  "trend": "Trend",
+  "legendForChart": "Legend for {chartType} chart"
```
sheila-semrush, line 3: writer approved "Chart legend" (without chart type)

ilyabrower: fixed

sheila-semrush (approved):

sheila-semrush:
@j-mnizhek Take a look at the updated docs pls. Here are some of the changes:

* Switched A11y and API tabs around. Didn't move the "Concept and code" tab, I thought it was probably an intentional exception, because charts are more complex and the tab is more important here. But if you think we should rename it to "Example" and move to the end, we can do that.

A11y tab:

* Structured like in other components. Moved some subsections around to fit in these categories.
* Moved summary examples to the "Concept and code" tab.
* Added attributes tables.
* Some content will probably change again soon, we're still going to change some things in the accessible summary dialog.

Concept and code tab:

* Reorganized the page, grouping some items together.
* Left the summary examples as they are for now, we'll need a separate task to turn them into live examples.
* Updated text about Skeleton.

Design tab:

* Removed the "Edge cases" title in the Tooltip section and bumped the children titles levels up. Vale complained about heading levels, and I think it's fine not to add grouping if it's only 2 sections, and this level doesn't appear in the TOC anyway.

j-mnizhek:
> * Switched A11y and API tabs around. Didn't move the "Concept and code" tab, I thought it was probably an intentional exception, because charts are more complex and the tab is more important here. But if you think we should rename it to "Example" and move to the end, we can do that.
Yep, initially developers named it that way to introduce the whole concept of the new charts built on D3.js. Actually I think, we can rename it to Example already, why not.

j-mnizhek (approved):
Looks and sound good!

PR #1667, opened 2024-09-25
[86c0dr0v2][utils] fixed focus in containers with focus-lock and iframes

ilyabrower:
## Motivation and Context
I've added focus-borders only for iframes outside focusTrap element.
For now - we will set focus in focus-lock container only if we have an element for that.

## How has this been tested?
Manually and with new browser-test

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/utils/src/use/useFocusLock.ts
```diff
@@ -136,7 +140,7 @@ const useFocusLockHook = (
   focusMaster = false,
   onFocusOut?: (event: Event) => void,
 ) => {
-  useFocusBorders(React, disabled);
+  useFocusBorders(React, trapRef, disabled);
```
msereniti, line 143: Let's add it as a third attribute (just to make focus lock merging safer if it will happen anywhere)

semcore/utils/src/use/useFocusLock.ts
```diff
@@ -166,7 +170,7 @@ const useFocusLockHook = (
         )
           return;
 
-        if (focusCameFrom) {
+        if (focusCameFrom && focusMovedTo) {
```
msereniti, line 173: What the point of this condition change?

msereniti (approved):
Approve but please change the order of arguments before mering

semcore/utils/src/use/useFocusLock.ts
ilyabrower: we need this to prevent the focus from remaining on the close button if the next focusable element is inside iframe

semcore/utils/src/use/useFocusLock.ts
ilyabrower: fixed

PR #1640, opened 2024-09-09
[86bztnr9z][carousel] navigatable carusel controls

msereniti:
## Motivation and Context

Basically, this PR reduced carousel usability from screen reader but now it will comply with formal requirments of a11y audits.

## How has this been tested?

Manually.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [x] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Nice improve.

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have updated the documentation accordingly or it's not required.
- [x] Unit tests are not broken.
- [x] I have added changelog note to corresponding `CHANGELOG.md` file with planned publish date.
- [ ] I have added new unit tests on added of fixed functionality.

PR #1669, opened 2024-09-25
[86bztnr9z][docs] made carousel examples a bit more accessible by wrapping images into tabpanel containers

msereniti:


j-mnizhek (approved):
Looks and sounds good 👍 

I see only that I cannot open the image in a fullscreen using keyboard (as I've notice here https://github.com/semrush/intergalactic/pull/1640#issuecomment-2352866189).

biome.json
```diff
@@ -88,7 +88,8 @@
       "a11y": {
         "useAnchorContent": "off",
         "noSvgWithoutTitle": "off",
-        "noNoninteractiveTabindex": "off"
+        "noNoninteractiveTabindex": "off",
+        "noNoninteractiveElementToInteractiveRole": "off"
```
ilyabrower, line 92: Maybe we shouldn't add this as global?

Moreover, let's add role button to the Carousel.Item ?

msereniti: > Maybe we shouldn't add this as global?

Makes sense

> Moreover, let's add role button to the Carousel.Item ?

It should has tabpanel role

ilyabrower (approved):

PR #1680, opened 2024-10-02
Update/table docs structure

j-mnizhek:
## Motivation and Context

We have a mess with DataTable examples in the documentation. I tried to structure all examples a little bit, I've added sections and scattered all the examples among them.

So, what do you think? Can it work better? _(or I will find and try some other options, I'm actually still concerned by the amount of time this page is loading)_

image

## How has this been tested?

Manually.

## Types of changes

- [x] Nice improve.

## Checklist:


website/docs/table-group/data-table/data-table-code.md
```diff
@@ -222,7 +222,33 @@ You can provide `data` property for `<DataTable.Row/>` . It isn't used in the co
 
 :::
 
-## Access to cells
+### Rows merging
```
sheila-semrush, line 225: I'd move both row and column merging to "Table cells" and rename it to "merging cells vertically" and "merging cells horizontally"

website/docs/table-group/data-table/data-table-code.md
```diff
@@ -375,45 +402,40 @@ Add a skeleton to the table by directly substituting it in the `data` or replaci
 
 :::
 
-## Columns merging
+## Specific cases
 
-Merge two or more columns by changing the table data and using `/` to combine column keys.
+### Accordion inside table
```
sheila-semrush, line 407: maybe create a section for all 3 accordion examples and call it "Accordion"?

website/docs/table-group/data-table/data-table-code.md
```diff
@@ -252,81 +278,80 @@ To apply properties to multiple table cells, define `<DataTable.Cell />` with th
 
 :::
 
-## Adding additional elements to table body
+### Custom footer cells
```
sheila-semrush, line 281: this seems like it can be renamed to "Custom footer row" and moved to the "Rows" section, because it's about a whole row in reality

website/docs/table-group/data-table/data-table-code.md
sheila-semrush, line 225: I'd rename it to "Row grouping"

website/docs/table-group/data-table/data-table-code.md
```diff
@@ -176,35 +174,37 @@ If fixed columns aren't visible in the following example, try reducing the windo
 
 :::
 
-## Multi-level header
+### Columns merging
```
sheila-semrush, line 177: this I'd actually move to "cells" and rename to "merging cells horizontally", because "columns merging" sounds like we merge entire columns together

website/docs/table-group/data-table/data-table-code.md
```diff
@@ -375,45 +402,40 @@ Add a skeleton to the table by directly substituting it in the `data` or replaci
 
 :::
 
-## Columns merging
+## Specific cases
```
sheila-semrush, line 405: if we move all the accordion examples out of here, we can rename this to "Custom elements in table" or something, then it will be much easier to find

sheila-semrush (commented):
I love it! I left a couple of suggestions to move some things around, but in general the structure seems very convenient.

website/docs/table-group/data-table/data-table-code.md
j-mnizhek: Not sure, since the last two with tables inside tables look like they are intended to show these specific cases with for nesting tables. But could be a good option.

I took the Accordion section out of the Specific cases and moved examples with tables at the top, so they will follow the Accordion section.

website/docs/table-group/data-table/data-table-code.md
j-mnizhek: Visually it looks like we are merging cells horizontally, but from the code side we are merging columns. I would leave the original title so as not to confuse developers when they are looking for a case of how to merge columns (but they will need to look for it in the section on cells).

Hmm, maybe it's better to update the example, so it wouldn't be confusing visually.

website/docs/table-group/data-table/data-table-code.md
j-mnizhek: Hmm, not sure about this. Wiil discuss it with Ilia a little bit.

website/docs/table-group/data-table/data-table-code.md
j-mnizhek: Discussed with Ilia, both "Columns merging" and "Columns grouping" are good. "Meging cells horizontally/vertically" don't suit for these cases.

website/docs/table-group/data-table/data-table-code.md
j-mnizhek: This example are more complicated, hehe. Here is Ilia's comment on how to name this:

> view after main table body which has  css variables for correct width for cell in flex row :smiley:

😁

website/docs/table-group/data-table/data-table-code.md
j-mnizhek: I think we can move two examples with table and tanle after we'll investigate a little whether they are necessary at all.

PR #1786, opened 2024-11-20
[UIK-2455][filter trigger] a11y

ilyabrower:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Nice improve.

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly or it's not required.
- [ ] Unit tests are not broken.
- [ ] I have added changelog note to corresponding `CHANGELOG.md` file with planned publish date.
- [ ] I have added new unit tests on added of fixed functionality.

PR #1812, opened 2024-11-28
[UIK-2683][docs] Updated Click Potential example and design page

sheila-semrush:
### A11y

- trigger must have role=combobox - fixed in #1786 
- Clear button must have a Hint - fixed in #1786 
- combobox value shouldn't include its name
- from/to inputs should have unique labels
- dialog label must be "Competitive density"
- InputRange must be a group with the "Custom range" label
- from/to inputs have invalid input role, and sometimes region role - fixed in #1821 
- from/to input values are read twice due to aria-live polite - discussed in #1821

### UX

- fix popper margins/size (were inconsistent with the design)
- fix from/to inputs behavior:
  - swap from/to only when both inputs are blured
  - if one of the inputs is empty, autofill it with min/max range value
  - if `from === to`, show single number in trigger
  - clear inputs on pressing the Clear button
  - first input should be auto focused on popper open

## How has this been tested?

Manually.

## Types of changes

- [x] Nice improve.

## Checklist:


sheila-semrush:
@ilyabrower can you review the example, pls

website/docs/filter-group/filter-cp-cd-cpc/examples/basic-example.tsx
```diff
@@ -7,30 +7,38 @@ import NeighborLocation from 'intergalactic/neighbor-location';
...
+    setTimeout(() => {
+      if (document.activeElement !== fromRef.current && document.activeElement !== toRef.current) {
+        const { from, to } = valueState;
+        if (from > to && to) {
+          changeValue({
+            from: Math.max(to, minRange),
+            to: Math.min(from, maxRange),
+          });
+        }
+        if ((!to || !from) && (to || from)) {
```
ilyabrower, line 32: Not sure about this, because `0` is false. And about condition on 26 line too (`&& to` could be invalid if `to=0`)

sheila-semrush: yeah you're right, minRange can be zero, fixed it

ilyabrower: I fixed it a little in another places. Everything is good

ilyabrower (approved):

website/docs/filter-group/filter-cp-cd-cpc/examples/basic-example.tsx
sheila-semrush: oops, thanks!

Valeria-Zimnitskaya (approved):

sheila-semrush:
@j-mnizhek can you look at the docs update?
The most significant change is that I clarified that InputRange values can update only when *both* inputs lose focus, and I implemented this logic in the example. You can also look at how the example works now.

j-mnizhek (approved):
Both documentation and examples UX looks good, so I'm approving 👍

PR #1821, opened 2024-12-02
[UIK-2749][input-number] removed unnecessary role changes

ilyabrower:
## How has this been tested?
Manually

## Types of changes

- [X] Nice improve.

## Checklist:


semcore/input-number/src/InputNumber.jsx
```diff
@@ -478,7 +470,7 @@ class Value extends Component {
         in the input, because after validation the value can change to the `min` or `max`
         if entered less than `min` or more than `max` */}
         <SValueHidden aria-live='polite' aria-atomic={true}>
```
sheila-semrush, line 472:
```suggestion
        <SValueHidden aria-live='polite' aria-atomic={true}>
```
hmm I hear the value twice after editing (both with Up/Down and direct input)
maybe we don't need this aria live?
it's also not very useful even in case of max/min auto correction, case:

- the max is 8, I enter 9 in "from" and press tab, "to" is selected
- I expect SR to read the "to" value, but SR first reads the corrected "from" value, and then the "to" value, and it's impossible to understand which is which

In general, I can see why we added aria-live, but it must be a little more sophisticated to be useful

ilyabrower: removed

sheila-semrush (approved):
Everything seems good!
Added Hints to the InputNumber example with +/- buttons.
There's no live messages when incrementing/decrementing in that example now, but we'll solve this properly in another task.

PR #1825, opened 2024-12-02
[UIK-2747][mini-chart] added role=img to all svg charts

ilyabrower:
## How has this been tested?
Manually

## Types of changes

- [X] Nice improve.

## Checklist:

### DIFF
--- a/semcore/mini-chart/CHANGELOG.md
+++ b/semcore/mini-chart/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [0.27.0] - 2024-12-02
+
+### Added
+
+- `role=img` to all svg charts.
+
 ## [0.26.0] - 2024-11-29
 
 ### Changed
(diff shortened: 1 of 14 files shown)


sheila-semrush (approved):
All good from my side!

@j-mnizhek can you review the doc update?
I added aria-hidden to all examples. Maybe we could add some sample data summaries to area/bar/line charts, but I think it's not really that necessary. Donuts more likely have to be aria-hidden anyway, because they usually have a completely identical text metric next to them.

UPD: just out of curiosity, I generated a summary for the first mini area chart. Here it is:

> Chart represents 1 time series: growing from 35 to 70, also strongly growing from 0 to 4, strongly declining from 4 to 6, strongly growing from 6 to 8, and strongly growing from 8 to 9. x represented from 0 to 9

It's a bit weird because it has numbers that actually aren't visible on the mini chart. I also had to invent the x axis. So I'm not sure if it makes sense to add summaries like this to our examples. Maybe if we remove all the numbers, but then it might be even weirder, like "...strongly growing in the beginning, then strongly declining, then strongly growing"

BTW, I don't think our summarizer is correct...

image

j-mnizhek:
> I added aria-hidden to all examples. Maybe we could add some sample data summaries to area/bar/line charts, but I think it's not really that necessary. Donuts more likely have to be aria-hidden anyway, because they usually have a completely identical text metric next to them.

Looks good for me also 👍 Thank you for updating the text, I was concerned about this and left a comment to update it along with this task. And now everything is updated already 😎

j-mnizhek (approved):
No additional comments from my side 👍

Valeria-Zimnitskaya (approved):

PR #1828, opened 2024-12-02
[UIK-2687][docs] Updated Volume filter example and docs

sheila-semrush:
### A11y

Still present, need to find a solution:
- problem navigating from bottom part to listbox UIK-2763
- both input and list item are focused, and pressing Enter is unpredictable
image

Fixed:
- problem with input roles - #1821 
- dialog and trigger must be labelled "Volume"
- listbox must be labelled "Presets"
- from/to inputs should have unique labels
- from/to inputs must be in a group named "Custom range"

### UX

All fixed:
- custom range must be cleared when a list option is applied
- no options must be selected in list when custom range is applied
- max value must be unlimited
- if "to" is empty, it must stay empty, and trigger must show "{from}" (e.g. "55")

## Types of changes

- [x] Nice improve.

## Checklist:


sheila-semrush:
@ilyabrower what do you think we should do to solve these 2 issues?

you're also welcome to review the code written so far

image

ilyabrower:
> @ilyabrower what do you think we should do to solve these 2 issues?

Fixed

Valeria-Zimnitskaya (approved):

j-mnizhek (approved):
Now example works excellent 😍 Nothing to add to this PR

PR #1851, opened 2024-12-09
[UIK-2771][d3-chart,illustration] added aria-hidden for svgs

ilyabrower:
## Types of changes

- [X] Nice improve.

## Checklist:


j-mnizhek (approved):
Works good with VO 😎
@Valeria-Zimnitskaya Lera, if you will have a possibility to check how charts are read with NVDA, it will be wonderful ❤️

j-mnizhek:
We’ve discussed the issue with charts that don’t have a wrapper with the `region` role and decided to set default `aria-label="Chart"` for svg inside charts, removing `aria-hidden="true"` from it. In this case, for charts with a full wrapper, the group with the chart name will be read first, followed by the SVG with the default name "Chart".

PR #1874, opened 2024-12-17
[UIK-2697][notice] a11y improvements

ilyabrower:
### Removed

- `aria-live` from all items.
- `aria-label` from "muted" theme.

### Added

- Log with recommendation to provide aria-labels that help distinguish different types of notices.

## How has this been tested?
Manually

## Types of changes

- [X] Nice improve.

## Checklist:

### DIFF
--- a/semcore/notice/src/Notice.jsx
+++ b/semcore/notice/src/Notice.jsx
@@ -36,6 +36,16 @@ class RootNotice extends Component {
     locale: 'en',
   };
 
+  constructor(props) {
+    super(props);
+
+    logger.warn(
+      !props['aria-label'],
+      'Provide aria-label to help identify the type of notification',
+      RootNotice.displayName,
+    );
+  }
+
   getLabelProps() {
     const { theme, resolveColor } = this.asProps;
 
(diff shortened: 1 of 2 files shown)


semcore/notice/src/Notice.jsx
```diff
@@ -36,6 +36,16 @@ class RootNotice extends Component {
     locale: 'en',
   };
 
+  constructor(props) {
+    super(props);
+
+    logger.warn(
```
sheila-semrush, line 42: actually, in the task I meant writing about this in the docs, but this is also neat
but
this shouldn't fire for the "muted" theme, and probably for the "critical" either. I think it would be reasonable to only show this warning for the themes that have the same "Notification" aria-label.

semcore/notice/src/Notice.jsx
```diff
@@ -36,6 +36,16 @@ class RootNotice extends Component {
...
+      !props['aria-label'],
+      'Provide aria-label to help identify the type of notification',
```
sheila-semrush, line 44:
```suggestion
      'Provide unique aria-label to help identify the type of notification',
```
probably better to clarify

semcore/notice/src/Notice.jsx
sheila-semrush: another option would be to keep showing this for all themes, but lower the level to info (or whatever is below warning)

sheila-semrush (approved):
Updated some examples and docs.
Overall it's fine, I added a couple of suggestions.

semcore/notice/src/Notice.jsx
ilyabrower: added check for themes

Valeria-Zimnitskaya (approved):

PR #1880, opened 2024-12-17
[website] changed hint to tooltip in disabled tapPanel(tabLine).Items

ilyabrower:
## Types of changes

- [X] Nice improve.

## Checklist:


sheila-semrush (approved):
One thing: the tooltip here will never be accessible for SR, because we set aria-describedby only when the tooltip is visible, and it will be never visible from keyboard, because the trigger is disabled. But I guess it'll do for this example, for the time being...

Valeria-Zimnitskaya (approved):

PR #1921, opened 2025-01-13
[docs] Updated FeedbackRating a11y docs

sheila-semrush:
## Motivation and Context

Forgot to do it together with the main task, so doing it in a separate PR.

1. Updated the a11y attributes list.
2. Changed the email error text in the example. It was "set correct email"—I think it was incorrect? Changed it to the same that's in the main feedback form.
3. Linter fixes. Left out 3 suggestions intentionally, one refers to the UI copy, and others I just felt that it's not worth complying with. I think we're justified to use "above/below" when we're describing UI design.

## Types of changes

- [x] Nice improve.

## Checklist:


j-mnizhek (approved):
Looks like everything is good 👍

PR #1942, opened 2025-01-23
[UIK-2735][docs] Labels and instructions

sheila-semrush:
## Motivation and Context

This PR adds missing labels and improves existing ones for various controls in our docs and examples.

Some highlights:

1. The only component change: fixed the name of a token in Button styles.
2. Dot: added aria-describedby to examples, but there still are a11y issues, I created a separate task.
3. ScrollArea: added aria-labels to scroll containers in examples, so when you focus a container it's clear what it is. Except for the first one, I can't add the label without expanding the component, and I wasn't sure it would be appropriate for the basic example.
4. Checkbox and Radio playgrounds: renamed "children" to "label".
5. InputTags playground: removed the text property because it didn't seem useful.
6. SpinContainer playground: changed the structure of inputs to be accessible when "loading" is off.
7. Spin playground: removed "centered" prop because its behavior didn't make sense, created a separate task. Also renamed the TextRight property.
8. Switch: added icon changing between on/off states logic in Playground and examples, kept only one addon in Playground and renamed it to "Label" for consistency with Radio & Checkbox. Created a task to add the "label" property instead addons.
9. Tag: rewrote playground to use TagContainer and Close correctly.
10. Charts: removed some useless Flex props from Playgrounds, renamed props to be clearer and shorter. Renamed "lines" and "bars" to "categories" (and vice versa) wherever needed (not everywhere, mostly in examples with legend). 
11. Box: removed the duplicate sandbox. I don't know why it was there.
12. Popper: Render functions example: added returning focus to 1st button after closing the popper, removed the duplicate sandbox. Controlled and uncontrolled example: replaced triggers with our Buttons.

## Types of changes

- [x] Nice improve.

## Checklist:


sheila-semrush:
@j-mnizhek in the end, I removed "centered" prop from Spin playground, it was too much trouble to make it make sense
I've created a task about that issue

j-mnizhek (approved):
Was checking if we really should move `justifyContent` prop from d3 chart playgrounds, and seems yep, we can live without it.

Everything other is good and clear!

Valeria-Zimnitskaya (approved):

PR #1951, opened 2025-01-31
[UIK-2891][website] Fixes for Transforming JSON widget

j-mnizhek:
## Motivation and Context

This PR fixes several accessibility issues we have now with Transforming JSON files widget on the Design tokens page.

### What I fixed:
- [x] Input has no visible focus
- [x] Description for the input wasn't connected with it (this text: "Upload files, uncompressed, less than 1 GB in size").
- [x] "Copy to clipboard" control wasn't accessible for the keyboard and SR
- [x] "Copy to clipboard" button wasn't connected to the title next to it
- [x] Heading hierarchy was wrong: h4 was following the h2.
- [x] "Replace file" button wasn't connected to some text that can give context what file should be replaced

**+ fixed some styles for this widget.**

### What should be fixed (I've described this ih the Jira task):
- [x] "Browse files" button isn't clickable by keyboard. Also hover state is snown by clicking Space, but by Enter — nothing.
- [x] The screen reader knows nothing about the error state or the processed code. You need to read this message. Perhaps we need aria-live="polite" on the block with the error and the processed code.
- [x] Since we changed span to our Button for the "Copy to clipboard" control near the processed code, now we have the redundunt tooltip (which comes from our inner Copy component for copy to clipboard functionality).
- [x] Nice to have (but can be made as a separate task). When I click "Replace file" button for Semantic tokens JSON file, somehow the input for Base tokens JSON file becomes active. Maybe it's intended behaviour, but I think we can make it better.
- [x] Nice to have (but can be made as a separate task). We should show the actual name of the file, not the default text, as in the native input.

## How has this been tested?

Manually and with VO.

## Types of changes

- [x] Nice improve.

## Checklist:


ilyabrower (approved):

j-mnizhek:
Hehehe, our button with visible text has tooltip with the same text. Could we hide the tooltip for this case?

image

PR #2037, opened 2025-03-18
[UIK-3330][docs] Added role img to Flags in input phone

sheila-semrush:
## Motivation and Context

Added `role=img` to InputPhone examples (2nd and 3d), so that flag in input/select trigger is announced as full country name.

Also edited aria-labels in the 2nd example, so that SR reads when focusing the input:
"Phone number, United States, ..."

(Flags in the Select Menu aren't announced, this is by design.)

## Types of changes

- [x] Nice improve.

## Checklist:

### DIFF
--- a/stories/components/input-phone/docs/examples/known_country_and_number_format.tsx
+++ b/stories/components/input-phone/docs/examples/known_country_and_number_format.tsx
@@ -18,57 +18,60 @@ const Demo = () => {
   const positionAfterFirstBracket = prefix.length + 2;
 
   const handleChange = (value: string, e: React.SyntheticEvent<HTMLInputElement>) => {
-      setPhoneNumber(value);
+    setPhoneNumber(value);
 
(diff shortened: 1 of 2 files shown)


sheila-semrush:
> Wouldn't it be convenient in the second example if the country's flag name were read along with the input label and text when the user comes to the input? My VO doesn't read it now. I was wondering whether it's worth linking the input value to the flag via aria-describedby, or would that be redundant? What do you all think?

Thank you for this comment, I had similar thoughts but didn't figure out how to solve this logically.
I decided to use both label and flag as aria-labelledby for the input, so it reads "Phone number, United States, edit text, ...". This seems more natural to me, with aria-describedby it read "Phone number, edit text, Playground, group... (pause) United States", that seemed confusing.

Valeria-Zimnitskaya (approved):

PR #2050, opened 2025-03-28
[docs] update AutoSuggest wording and fix bulkTextarea svg name again

j-mnizhek:
## Motivation and Context

Just a small wording update to the AutoSuggest pattern description.
Reason: a designer misunderstood that these patterns apply only to dropdown menus and selects.

I also updated the name of the BulkTextarea SVG for the Component showcase page. Again (for some reason it went back to all lowercase).

## How has this been tested?

Manually.

## Types of changes

- [x] Nice improve.

## Checklist:


website/docs/components/auto-suggest/auto-suggest.md
```diff
@@ -5,7 +5,7 @@ tabs: Design('auto-suggest'), A11y('auto-suggest-a11y'), Example('auto-suggest-c
 
 ## Description
 
-**AutoSuggest**, **AutoComplete** and **Combobox** are different patterns for as-you-type suggestions for search with values filtering from the list.
+**AutoSuggest**, **AutoComplete** and **Combobox** are patterns for providing as-you-type suggestions by filtering options within dropdowns and selects.
```
sheila-semrush, line 8: filtering is not relevant for combobox... maybe beter not to mention filtering in the general definition?

sheila-semrush: what was the initial problem? designer didn't realize these patterns have a text input in them? 
if yes, then maybe better highlight the part about input, like

> ...are patterns for providing a list of suggestions when typing in a text input

website/docs/components/auto-suggest/auto-suggest.md
```diff
@@ -5,7 +5,9 @@ tabs: Design('auto-suggest'), A11y('auto-suggest-a11y'), Example('auto-suggest-c
 
 ## Description
 
-**AutoSuggest**, **AutoComplete** and **Combobox** are different patterns for as-you-type suggestions for search with values filtering from the list.
+**AutoSuggest** and **AutoComplete** are patterns for providing as-you-type suggestions by filtering options within dropdowns and selects.
+
+**Combobox** is a pattern that provides a user with a predefined list of options.
```
sheila-semrush, line 10: should we mention typing here, so people don't think that combobox is the same as select in our case?

j-mnizhek: Hmmmmm, I was thinking whether to add it, but didn't add because in many cases (and even in our example) you don't need to type something to see the list

sheila-semrush: it shouldn't necessarily seem as if user needs to type. Like for example something like this in the end of the sentence: "...to help with filling out an input."

sheila-semrush (approved):

PR #2284, opened 2025-06-20
[UIK-3441][modal] allow ESC to close modal even without focus inside

slizhevskyv-semrush:
## Motivation and Context

Previously, if no element inside the Modal has focus, pressing the ESC key would not close the modal.
The fix ensures that the ESC key closes the modal regardless of the current focus state, as long as the modal is open.

## How has this been tested?

I've added a browser test.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/modal/__tests__/modal.browser-test.tsx
+++ b/semcore/modal/__tests__/modal.browser-test.tsx
@@ -51,6 +51,16 @@ test.describe('Modal interactions', () => {
       await expect(modal).toHaveCount(0);
       await expect(trigger).toBeFocused();
     });
+
+    await test.step('Verify modal closed when you firstly click on a modal and then pressing ESC', async () => {
+      await trigger.click();
+      await modal.waitFor({ state: 'attached' });
+
+      await modal.click();
+      await page.keyboard.press('Escape');
+
+      await expect(modal).toHaveCount(0);
+    });
   });
 
   test('Verify basic usage mouse interactions', async ({ page, browserName }) => {
--- a/semcore/modal/CHANGELOG.md
+++ b/semcore/modal/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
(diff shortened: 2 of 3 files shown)


semcore/modal/__tests__/modal.browser-test.tsx
```diff
@@ -51,6 +51,16 @@ test.describe('Modal interactions', () => {
       await expect(modal).toHaveCount(0);
       await expect(trigger).toBeFocused();
     });
+
+    await test.step('Verify modal closed when you firstly click on a modal and then pressing ESC', async () => {
+      await trigger.click();
+      await modal.waitFor({ state: 'attached' });
+
+      await modal.click();
```
ilyabrower, line 59: It should work even without this click

slizhevskyv-semrush: It's a step to reproduce the issue. It works without click, that's true :)

ilyabrower: With this click you verify only that modal has tabIndex={-1} as in your fix (and the modal will have focus after click).
Without it, you will also check that the focus will be set after modal opens.

semcore/color-picker/__tests__/color-picker.browser-test.tsx
```diff
@@ -320,20 +320,23 @@ test.describe('Color-picker', () => {
     await test.step('Verify hover on No background color ', async () => {
       const items = colorPoppers.first().getByRole('option');
       await items.first().hover();
-      await new Promise((resolve) => setTimeout(resolve, 300));
+      // await new Promise((resolve) => setTimeout(resolve, 300));
```
ilyabrower: let's remove this row

ilyabrower (dismissed):

semcore/color-picker/__tests__/color-picker.browser-test.tsx
slizhevskyv-semrush: updated :)

ilyabrower (approved):

Valeria-Zimnitskaya (approved):

PR #2298, opened 2025-06-27
[UIK-3523][website] added unique titles for Loom frames

sheila-semrush:
## Motivation and Context

Changed our custom `loom_video` plugin to insert specified title into the iframe.

Now loom videos should be embedded like this:
_(title is optional, default one is "video")_
imagen

For existing videos on our website, I copied titles from the videos themselves, as per [Axe's recommendation](https://dequeuniversity.com/rules/axe/4.10/frame-title-unique):

> best practice is to give the enclosed document a title element with content identical to the title attribute. Some screen readers will replace the contents of the title attribute on the frame with the contents of the title element inside the frame. As a result, it’s safest and most accessible to have the same text in both locations.

## Types of changes

- [x] Nice improve.

## Checklist:

### DIFF
--- a/website/docs/.vitepress/renderLoomVideo.ts
+++ b/website/docs/.vitepress/renderLoomVideo.ts
@@ -1,17 +1,13 @@
 export const renderLoomVideo = (tokenList: any[], index: number) => {
   const renderFunc = (tokens: any[], idx: number) => {
     const token = tokens[idx];
-    if (token.type === 'container_loom_video_open') {
-      let url = token.info.split(' ')[2];
-      if (!url) return [];
+    if (token.nesting === 1) {
+      const title = token.info.replace('loom_video', '').trim() || 'video';
+      const url = tokens[idx + 2].content;
 
-      if (url.startsWith('https://www.loom.com/share/')) {
-        url = `https://www.loom.com/embed/${url.substring('https://www.loom.com/share/'.length)}`;
-      }
-
-      return `<div class="embedded-video-container"><iframe src="${url}" frameborder='0' webkitAllowFullScreen mozAllowFullScreen allowFullScreen class="embedded-video-iframe" title='video' /></div>`;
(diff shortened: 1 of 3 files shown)


website/docs/.vitepress/renderLoomVideo.ts
```diff
@@ -1,17 +1,13 @@
...
+    if (token.nesting === 1) {
+      const title = token.info.replace('loom_video', '').trim() || 'video';
+      const url = tokens[idx + 2].content;
 
-      if (url.startsWith('https://www.loom.com/share/')) {
-        url = `https://www.loom.com/embed/${url.substring('https://www.loom.com/share/'.length)}`;
-      }
-
-      return `<div class="embedded-video-container"><iframe src="${url}" frameborder='0' webkitAllowFullScreen mozAllowFullScreen allowFullScreen class="embedded-video-iframe" title='video' /></div>`;
+      return `<div class="embedded-video-container"><iframe src='${url}' frameborder='0' webkitAllowFullScreen mozAllowFullScreen allowFullScreen class="embedded-video-iframe" title='${title}'>`;
```
sheila-semrush, line 8: @ilyabrower this wraps `<iframe>...</iframe>` around the original link from the source code, because we only manipulate the opening and the closing tags, not the content inside. It's replaced by the frame content when it's loaded, but we get a warning:
`<p> cannot be child of <iframe>, according to HTML specifications. This can cause hydration errors or potentially disrupt future functionality.`
I don't know how critical it is. Should we do something with the inner content, or can we leave it as is for now?

j-mnizhek (approved):
👍

website/docs/.vitepress/renderLoomVideo.ts
ilyabrower: this is a strange behaviour of markdown-it, we can skip this error, on the page there is correct iframe without any unnecessary tags.

Valeria-Zimnitskaya (approved):

PR #2302, opened 2025-07-01
[UIK-2652][input] preserve focus when closing with Space key

slizhevskyv-semrush:
## Motivation and Context

When the user activates the Close button using the Space key, focus is lost instead of being preserved or redirected.
This breaks expected keyboard navigation.

## How has this been tested?

I've added browser tests to handle ESC and Space cases.

## Types of changes

- [x] Nice improve.

## Checklist:


semcore/input/__tests__/input.browser-test.tsx
```diff
@@ -39,6 +39,62 @@ test.describe('Input', () => {
...
+    await expect(input).not.toBeFocused();
+    await expect(addonButton).toBeFocused();
+
+    await page.keyboard.press('Enter');
+
+    await expect(input).toBeFocused();
+    await expect(addonButton).toBeHidden();
+    expect(await input.inputValue()).toBe('');
+
+    await expect(page).toHaveScreenshot();
```
ilyabrower: I'm not sure, that we need a screenshot here... what do you think?

semcore/input/__tests__/input.browser-test.tsx
```diff
@@ -39,6 +39,62 @@ test.describe('Input', () => {
...
+    await expect(input).not.toBeFocused();
+    await expect(addonButton).toBeFocused();
+
+    await page.keyboard.press('Space');
+
+    await expect(input).toBeFocused();
+    await expect(addonButton).toBeHidden();
+    expect(await input.inputValue()).toBe('');
+
+    await expect(page).toHaveScreenshot();
```
ilyabrower: and here

semcore/input/src/Input.jsx
```diff
@@ -23,8 +23,8 @@ class Input extends Component {
     this.inputRef.current?.focus();
   };
 
-  handleKeyDown = (event) => {
-    if (event.key === 'Enter') {
+  handleKeyUpAndDown = (event) => {
```
ilyabrower: why not only key up?
maybe we should add this handler to addon only?

semcore/input/__tests__/input.browser-test.tsx
slizhevskyv-semrush: yeah, kind of agree, we have already verified that the focus is set before. Will remote it.

semcore/input/src/Input.jsx
slizhevskyv-semrush: yes, it would be a great idea to apply it only for addons.
there is weird behaviour for `Enter` and `Space` when they pressed. The focus for `Enter` returns back to the input only in `handleKeyDown`, while `handleKeyUp` works only for `Space` :)
I suppose that the reason is how browser handles event chain for those keys (I mean the order at which `onClick` is called)

slizhevskyv-semrush: `Space`:
The focus does not have enough time to move to `document.body` at `handleKeyDown`. If I add more time,e.g `200` instead of `10` then it works as a charm :) But it's weird :)

Valeria-Zimnitskaya (dismissed):

semcore/input/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [16.0.5] - 2025-07-07
+
+### Fixed
+
+- Focus is lost when user activates Close button with Space key.
```
j-mnizhek:
```suggestion
- Focus is lost when user activates Close button with `Space` key.
```

Just a small update to match the formatting of the keyboard keys 🎹

j-mnizhek (approved):
Everything works as expected 😎 Left only one tiny tiny suggestion

slizhevskyv-semrush:
> Everything works as expected 😎 Left only one tiny tiny suggestion

Thanks, updated!

PR #2308, opened 2025-07-03
Draft: Add Intergalactic CSS layer (for tailwindcss compatibility)

PavelKoroteev:
Wrap every intergalactic style into the layers. In this case, any unlayered author-defined classes will have more specificity.

## Motivation and Context

While using custom classes with intergalactic, it's sometimes issues with styles cascading (the intergalactic classes and author classes have the same specificity, therefore total styles depend on the builder).

When authors define their classes, they have the same specificity as intergalactic; therefore, they will not have any effect when they appear before intergalactic ones. (i.e., using tailwindcss with intergalactic).

## How has this been tested?

There's no any testing, only draft for consideration.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/core/src/styled/sstyled.tsx
+++ b/semcore/core/src/styled/sstyled.tsx
@@ -14,7 +14,7 @@ const getStyles = () => ({
   get css() {
     let serverStyles = '';
     for (const id in serverMap) {
-      serverStyles += `<style type='text/css' id='${id}'>${serverMap[id]}</style>`;
+      serverStyles += `<style type='text/css' id='${id}'>@layer semcore {${serverMap[id]}}</style>`
     }
     return serverStyles;
   },
@@ -53,7 +53,7 @@ function insert(code: any, hash: any) {
     container.appendChild(css);
   }
 
-  css.innerHTML = code;
+  css.innerHTML = `@layer semcore {${code}}`;
 }
 
 function merge(s1 = {}, s2 = {}) {


semcore/core/src/styled/sstyled.tsx
```diff
@@ -12,9 +12,10 @@ const serverMap = reshadowGetStyles().map as any;
 const getStyles = () => ({
   map: serverMap,
   get css() {
-    let serverStyles = '';
+    // Actually, we need to define layers order at first in html somehow.
+    let serverStyles = '<style type="text/css">@layer theme, base, semcore, components, utilities;</style>';
```
PavelKoroteev, line 16: This layers statement should be on the consumer side.

semcore/core/src/styled/sstyled.tsx
```diff
@@ -12,9 +12,10 @@ const serverMap = reshadowGetStyles().map as any;
...
     for (const id in serverMap) {
-      serverStyles += `<style type='text/css' id='${id}'>${serverMap[id]}</style>`;
+      serverStyles += `<style type='text/css' id='${id}'>@layer semcore {${serverMap[id]}}</style>`
```
PavelKoroteev, line 18: The best solution would be `@import semcore.css layer(semcore)` inside some consumer bundler loader.

PR #2318, opened 2025-07-07
Small fixes after Feature highlight release

sheila-semrush:
## Motivation and Context

1. Added FH tokens to dark.json. They aren't supported by our website, created a separate task for this.
2. Added missing redirects for our website.
3. Removed unnecessary import from Select example.

Question: I added a changelog record, but I'm not sure about it. We didn't add one after initially adding FH tokens, and I'm also unsure about versioning (our rules differ from the official semantic versioning).

## Types of changes

- [x] Nice improve.

## Checklist:


semcore/core/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [16.1.0] - 2025-07-07
```
Valeria-Zimnitskaya: As I understand it, the component version must be incremented by a patch version (since we have an 'added' entry in the changelog), so it might be 16.0.4
@slizhevskyv-semrush what do you think?

sheila-semrush: thing is, in the url above you'll read that adding backward compatible functionality is minor, and patch is for fixing bugs
although, in this case this could have been a bug... I think it's not that important in this case, I'll change it to patch

semcore/core/CHANGELOG.md
sheila-semrush, line 1: oh wait, we also need to add a page with FH changelog, I'll do it shortly

sheila-semrush: added

semcore/core/CHANGELOG.md
slizhevskyv-semrush: agree with @sheila-semrush. Thanks!

PR #2319, opened 2025-07-08
[UIK-3901][data-table] fixed row hover effect for empty state

slizhevskyv-semrush:
## Motivation and Context

The row hover effect is still applied for empty state. It fixed the hover background for that empty row.

## How has this been tested?

I've added a browser test for that.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/data-table/src/components/Body/Row.types.ts
+++ b/semcore/data-table/src/components/Body/Row.types.ts
@@ -22,6 +22,7 @@ export type DataTableRowProps = {
   isAccordionRow?: DataTableCellProps['isAccordionRow'];
   animationExpand?: DataTableCellProps['animationExpand'];
   accordionRowIndex?: DataTableCellProps['accordionRowIndex'];
+  isEmptyRow?: boolean;
 };
 
 export type RowPropsInner = JSX.IntrinsicElements['div'] & {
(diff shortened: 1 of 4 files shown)


semcore/data-table/src/components/Body/Row.types.ts
```diff
@@ -22,6 +22,7 @@ export type DataTableRowProps = {
   isAccordionRow?: DataTableCellProps['isAccordionRow'];
   animationExpand?: DataTableCellProps['animationExpand'];
   accordionRowIndex?: DataTableCellProps['accordionRowIndex'];
+  isEmptyRow?: boolean;
```
sheila-semrush: what if we call it something like `isNonInteractive`, as more universal? (in case we want to use non-hoverable rows for something else)

semcore/data-table/src/components/Body/style.shadow.css
```diff
@@ -47,6 +47,10 @@ SCollapseRow > SCellWrapper > SCell:not([theme]), SRow[isAccordionRow] > SCellWr
 
 /* we need a media query here because of the postcssHoverMediaFeature plugin. it doesn't handle this type of selectors correctly */
 @media (hover: hover) {
+  SRow[isEmptyRow] > SCellWrapper > SCell {
+    background-color: unset !important;
```
sheila-semrush: it shouldn't be unset, that makes it transparent, it should be `--intergalactic-bg-primary-neutral`
wouldn't it be better instead of adding a new rule for this, to add a `:not([isEmptyRow])` to the default hover rule?

semcore/data-table/src/components/Body/Row.types.ts
slizhevskyv-semrush: yeah, sounds more universal :)

semcore/data-table/src/components/Body/style.shadow.css
slizhevskyv-semrush: it sounds reasonably, thanks!

sheila-semrush (approved):

ilyabrower (approved):

PR #2321, opened 2025-07-08
UIK-3784/product-head update tests

Valeria-Zimnitskaya:
## Types of changes

- [x] Nice improve.

## Checklist:


semcore/product-head/__tests__/product-head.browser-test.tsx
```diff
@@ -0,0 +1,80 @@
+import { e2eStandToHtml } from '@semcore/testing-utils/e2e-stand';
+import { expect, test } from '@semcore/testing-utils/playwright';
+
+test.describe('Visual', () => {
+  const breakpoints = [
+    { width: 320 },
+    { width: 768 },
+    { width: 1920 },
```
ilyabrower, line 9: I think we should add something like `1200` and `1400 / 1600` too

Valeria-Zimnitskaya: I'm not sure we should add both. I added one, and I think that's enough
I also renamed the test - removed the word "each" since it might create expectations that don't align with the actual idea of the test
I wanted to use some boundary widths (small and large, based on our stats) and one, or at most two, in the middle

PR #2322, opened 2025-07-09
[UIK-3887][data-table] fixed incorrect 'Select all' checkbox behavior with pagination

slizhevskyv-semrush:
## Motivation and Context

`Select all` checkbox does not reflect correct selection state when paging is used. The logic is updated to account for paginated data.

## How has this been tested?

I've added a browser test.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/data-table/src/components/DataTable/DataTable.tsx
+++ b/semcore/data-table/src/components/DataTable/DataTable.tsx
@@ -91,6 +91,8 @@ class DataTableRoot<D extends DataTableData> extends Component<
 
   private headerNodesMap = new Map();
 
+  private persistedSelectedRows = new Set<UniqRowKey>();
+
   constructor(props: DataTableProps<D>) {
     super(props);
 
@@ -119,11 +121,15 @@ class DataTableRoot<D extends DataTableData> extends Component<
   }
 
   componentDidMount() {
-    const { headerProps, loading } = this.asProps;
+    const { headerProps, loading, selectedRows } = this.asProps;
     if ((headerProps?.sticky && !headerProps.h) || loading || this.columns.some((c) => c.fixed)) {
       requestAnimationFrame(() => {
         this.forceUpdate();
       });
+    }
+
+    if (selectedRows && selectedRows.length > 0) {
+      selectedRows.forEach(this.persistedSelectedRows.add, this.persistedSelectedRows);
     }
   }
 
@@ -145,6 +151,7 @@ class DataTableRoot<D extends DataTableData> extends Component<
         this.setSelectAllMessage(true);
       } else if (prevProps.selectedRows.length > 0 && selectedRows.length === 0) {
         this.setSelectAllMessage(false);
+        this.persistedSelectedRows.clear();
       }
     }
   }
@@ -228,14 +235,14 @@ class DataTableRoot<D extends DataTableData> extends Component<
       gridTemplateAreas,
       sideIndents,
       totalRows: this.totalRows,
-      selectedRows: selectedRows,
+      selectedRows,
+      persistedSelectedRows: this.persistedSelectedRows,
+      flatRows: this.flatRows,
       onChangeSelectAll: (value, e) => {
-        if (value === false) {
-          onSelectedRowsChange?.([], e);
-        } else {
-          const selectedRows = this.flatRows.map((row) => row[UNIQ_ROW_KEY]);
-          onSelectedRowsChange?.(selectedRows, e);
-        }
+        const selectedRows = this.flatRows.map((row) => row[UNIQ_ROW_KEY]);
+        selectedRows.forEach(value ? this.persistedSelectedRows.add : this.persistedSelectedRows.delete, this.persistedSelectedRows);
+
+        onSelectedRowsChange?.([...this.persistedSelectedRows], e);
       },
       getFixedStyle: this.getFixedStyle,
       onCellClick: this.handleCellClick,
@@ -312,21 +319,15 @@ class DataTableRoot<D extends DataTableData> extends Component<
   ) => {
     const { selectedRows, onSelectedRowsChange } = this.asProps;
 
-    if (selectedRows && onSelectedRowsChange) {
-      const newSelectedRows = new Set(selectedRows);
+    if (!selectedRows || !onSelectedRowsChange) return;
 
-      if (isSelected && !newSelectedRows.has(row[UNIQ_ROW_KEY])) {
-        newSelectedRows.add(row[UNIQ_ROW_KEY]);
-      } else if (!isSelected && newSelectedRows.has(row[UNIQ_ROW_KEY])) {
-        newSelectedRows.delete(row[UNIQ_ROW_KEY]);
-      }
-
-      onSelectedRowsChange([...newSelectedRows], event, {
-        selectedRowIndex,
-        isSelected,
-        row,
-      });
+    if (this.persistedSelectedRows.has(row[UNIQ_ROW_KEY])) {
+      this.persistedSelectedRows.delete(row[UNIQ_ROW_KEY]);
+    } else {
+      this.persistedSelectedRows.add(row[UNIQ_ROW_KEY]);
     }
+
+    onSelectedRowsChange([...this.persistedSelectedRows], event, { selectedRowIndex, isSelected, row });
   };
 
   setSelectAllMessage = (selectedAll: boolean) => {
--- a/semcore/data-table/src/components/Head/Head.tsx
+++ b/semcore/data-table/src/components/Head/Head.tsx
@@ -10,7 +10,7 @@ import { Group } from './Group';
 import type { DataTableGroupProps } from './Group.type';
 import type { DataTableHeadProps, HeadPropsInner } from './Head.types';
 import style from './style.shadow.css';
-import { DataTableInternal, SELECT_ALL } from '../DataTable/DataTable';
+import { DataTableInternal, SELECT_ALL, UNIQ_ROW_KEY } from '../DataTable/DataTable';
 import type { DataTableData } from '../DataTable/DataTable.types';
 
 class HeadRoot<D extends DataTableData> extends Component<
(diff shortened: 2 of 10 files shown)


stories/components/data-table/docs/examples/checkbox-in-table.tsx
```diff
@@ -28,70 +28,74 @@ const Demo = () => {
     return () => clearTimeout(timer);
   }, [ariaMessage]);
 
-  const limit = 5;
-  const tableData = data.slice(currentPage * limit, currentPage * limit + limit);
+  const perPage = [5, 3, 7, 10];
```
Valeria-Zimnitskaya: @slizhevskyv-semrush I've modified the example to display a different number of rows on each page. Could you please review the code to check if everything looks correct? Functionally, it works.

slizhevskyv-semrush: yeah, that looks good to me! Thanks!

sheila-semrush: better not do it here, in a documentation example we should have the same number of rows on each page (except the last one). If we need this for testing, better do it in a separate test example

stories/components/data-table/docs/examples/checkbox-in-table.tsx
```diff
@@ -37,218 +28,103 @@ const Demo = () => {
...
-      <Collapse
-        visible={!!selectedRows.length}
-        duration={150}
-        style={{ position: 'sticky', top: 0, zIndex: 2 }}
+    <>
+      <Box
+        tabIndex={-1}
+        wMax={800}
+        h='100%'
+        hMax={250}
```
sheila-semrush: let's make the height of this box static, so that Pagination doesn't move when switching to a page with less items

stories/components/data-table/docs/examples/checkbox-in-table.tsx
```diff
@@ -1,24 +1,15 @@
 import { Box, Flex, Collapse, ScreenReaderOnly } from '@semcore/base-components';
 import Button from '@semcore/button';
-import { sstyled } from '@semcore/core';
 import { DataTable } from '@semcore/data-table';
+import Pagination from '@semcore/pagination';
 import { Text } from '@semcore/typography';
 import React from 'react';
 
-const style = sstyled.css`
-  SDataTable > [aria-rowindex='1'] > div {
```
sheila-semrush, line 9: discussed that we need to add these styles to the component itself

stories/components/data-table/docs/examples/checkbox-in-table.tsx
Valeria-Zimnitskaya: @sheila-semrush got it, i'll move it to tests folder

Valeria-Zimnitskaya: @sheila-semrush  moved my variant to the tests folder and reverted the updates I made in the docs folder for the checkbox example. Now rows count is back to its original state (without my changes)

stories/components/data-table/docs/examples/checkbox-in-table.tsx
slizhevskyv-semrush: is it still relevant, since we have a constant value for a limit?

semcore/data-table/src/components/Head/style.shadow.css
```diff
@@ -2,6 +2,10 @@ SHead, SGroupContainer {
   display: contents;
 }
 
+SHead > SColumn {
```
slizhevskyv-semrush: @sheila-semrush i've moved the transition to the component itself. Just for now I left 150ms as a constant. In case of any requests it could be configurable through the props.

stories/components/data-table/docs/examples/checkbox-in-table.tsx
sheila-semrush: it works only if all rows are the same height and the total number of rows divides by the number on page, but in real examples the last page can contain less rows, so it will jump up on the last page

semcore/data-table/src/components/Head/style.shadow.css
```diff
@@ -2,6 +2,10 @@ SHead, SGroupContainer {
...
+  transition: top 150ms ease-out;
```
sheila-semrush: just leaving a reminder of what we've discussed:
- replace the 150ms with our `duration-fast` var
- in firefox, collapse has a delay, so there's a gap between the TH and the action bar, it would be nice if we could do something about it (if we can't solve the root problem, we could add a 80ms transition-delay just for FF)

stories/components/data-table/docs/examples/checkbox-in-table.tsx
```diff
@@ -37,218 +28,101 @@ const Demo = () => {
...
+  { id: '16', keyword: 'best buy coupons', kd: '48', cpc: '$3.70', vol: '17,100' },
+  { id: '17', keyword: 'walmart near me', kd: '60.1', cpc: '$0.95', vol: '50,000' },
+  { id: '18', keyword: 'netflix gift card', kd: '39', cpc: '$2.20', vol: '8,900' },
+  { id: '19', keyword: 'www.apple.com', kd: '71', cpc: '$6.90', vol: 'n/a' },
+  { id: '20', keyword: 'nike running shoes men', kd: '44', cpc: '$3.60', vol: '21,700' },
+  { id: '21', keyword: 'download spotify premium', kd: '58', cpc: '$4.75', vol: '26,800' },
+  { id: '22', keyword: 'buy dell laptop', kd: '53.1', cpc: '$5.40', vol: '19,600' },
+  { id: '23', keyword: 'gap kids sale', kd: '34', cpc: '$1.10', vol: '5,300' },
+  { id: '24', keyword: 'www.shein.com', kd: '69.8', cpc: '$3.95', vol: 'n/a' },
+  { id: '25', keyword: 'mcdonalds delivery app', kd: '29', cpc: '$2.05', vol: '14,200' },
```
sheila-semrush: I think we should remove a couple of last items, so that it's more obvious that pagination shouldn't move even if last page has less items

semcore/data-table/src/components/Head/style.shadow.css
slizhevskyv-semrush: There is only one working (stable) way to do so (add smth for a specific browser), but I think it's a little bit overhead. I just add a delay for all browsers.

stories/components/data-table/docs/examples/checkbox-in-table.tsx
slizhevskyv-semrush: good point, thanks!

sheila-semrush (approved):

stories/components/data-table/docs/examples/checkbox-in-table.tsx
sheila-semrush: I decreased the height a little bit, it looks ugly but it demonstrates the sticky header behavior better

semcore/data-table/src/components/DataTable/DataTable.tsx
```diff
@@ -119,12 +121,16 @@ class DataTableRoot<D extends DataTableData> extends Component<
...
-    const { headerProps, loading } = this.asProps;
+    const { headerProps, loading, selectedRows } = this.asProps;
     if ((headerProps?.sticky && !headerProps.h) || loading || this.columns.some((c) => c.fixed)) {
       requestAnimationFrame(() => {
         this.forceUpdate();
       });
     }
+
+    if (selectedRows && selectedRows.length > 0) {
+      selectedRows.forEach(this.persistedSelectedRows.add, this.persistedSelectedRows);
```
ilyabrower: I'm not sure that we need second argument here... because we don't need `this` in `add` method, no?

semcore/data-table/src/components/Head/Head.tsx
```diff
@@ -103,25 +103,36 @@ class HeadRoot<D extends DataTableData> extends Component<
     this.asProps.onChangeSelectAll?.(value, event);
   };
 
+  get areAllRowsSelected() {
+    const { selectedRows, flatRows } = this.asProps;
+
+    return flatRows.every((row) => selectedRows?.includes(row[UNIQ_ROW_KEY]));
```
ilyabrower, line 109: why do you check `selectedRows` instead of `persistedSelectedRows`?

semcore/data-table/src/components/Head/Head.tsx
```diff
@@ -103,25 +103,36 @@ class HeadRoot<D extends DataTableData> extends Component<
...
 
-    const checked = selectedRows && selectedRows.length === totalRows && totalRows > 0;
-    const indeterminate = selectedRows && selectedRows.length > 0 && !checked;
+    const areAllRowsSelected = this.areAllRowsSelected;
+    const indeterminate = this.isIndeterminate && !areAllRowsSelected;
 
     return sstyled(styles)(
       <>
-        <SHead render={Box} role='row' aria-rowindex={1}>
+        <SHead render={Box} role='row' aria-rowindex={1} use:sticky={sticky}>
```
ilyabrower: why do you need `use:` here?

semcore/data-table/src/components/Head/style.shadow.css
```diff
@@ -2,6 +2,10 @@ SHead, SGroupContainer {
   display: contents;
 }
 
+SHead > SColumn {
+  transition: top var(--intergalactic-duration-fast, 200) ease-out;
```
ilyabrower: I'm not sure that we need this transition for all headers... let's discuss why you added this)

stories/components/data-table/tests/data-table-states.stories.tsx
```diff
@@ -1,6 +1,8 @@
 import { DataTable } from '@semcore/data-table';
 import type { Meta, StoryObj } from '@storybook/react-vite';
 
+import CheckboxInTableDemo from './examples/table-states-tests/checkbox-in-table.tsx';
+import CheckboxWithPaginationAndDiffRowsAmountExample from './examples/table-states-tests/checkbox-pagination-diff-rows-count.tsx';
```
ilyabrower: I think, we should test real example and we don't need to use these two

semcore/data-table/src/components/DataTable/DataTable.tsx
slizhevskyv-semrush: we need this for `add` function of set. Since we aren't calling `add` with `add()`, we just pass a ref to this function.

semcore/data-table/src/components/Head/Head.tsx
slizhevskyv-semrush: it's needed to preserve valid behaviour for each page. If there is nothing on a page, `Select all` chackbox has to be unchecked even if there were checked items before (on the page before).

stories/components/data-table/tests/data-table-states.stories.tsx
Valeria-Zimnitskaya: i added and need this two to have in tests different cases. not only one from documentation. they are not 'unreal'

semcore/data-table/src/components/Head/Head.tsx
slizhevskyv-semrush: the better question here is wtf is `sticky` doing here? 😆

slizhevskyv-semrush: yeah, `use` isn't needed there :)

semcore/data-table/src/components/Head/style.shadow.css
slizhevskyv-semrush: I added this transition because from my perspective adding styles like this https://github.com/semrush/intergalactic/pull/2322/files#diff-271849f510a0f54fdbcf6ef23def97e04b097f4508eec6e232452f5c7f2d66d7L8
is a hack.

That is why we decided to move it to the component itself. Yeah, probably it should be configured by props, whether you would like this animation or not. But, yeah, we could add it as well if that makes sense.

semcore/data-table/src/components/Head/Head.tsx
slizhevskyv-semrush: updated

semcore/data-table/src/components/Head/Head.tsx
```diff
@@ -125,7 +125,7 @@ class HeadRoot<D extends DataTableData> extends Component<
 
     return sstyled(styles)(
       <>
-        <SHead render={Box} role='row' aria-rowindex={1} use:sticky={sticky}>
+        <SHead render={Box} role='row' aria-rowindex={1} sticky={sticky}>
```
ilyabrower: hm... you don't need sticky at all, it will work because of `const SHead = Root;`

ilyabrower (approved):

stories/components/data-table/tests/data-table-states.stories.tsx
Valeria-Zimnitskaya: Left one example and updated the tests accordingly

semcore/data-table/src/components/Head/Head.tsx
```diff
@@ -130,7 +130,7 @@ class HeadRoot<D extends DataTableData> extends Component<
           role='row'
           aria-rowindex={1}
           sticky={sticky}
-          animationDuration={animationDuration ? `${animationDuration}ms` : undefined}
+          use:animationDuration={animationDuration ? `${animationDuration}ms` : undefined}
```
ilyabrower, line 133: why do you use `use:` here?

slizhevskyv-semrush: without `use` applied only numeric value -> 200
with `use` applied string -> `200ms` 
🤷‍♂️

ilyabrower: oh.... yes, because of naming...

ilyabrower (dismissed):

Valeria-Zimnitskaya (dismissed):

PR #2325, opened 2025-07-10
[UIK-3921][bulk-textarea] fixed styling in types o_O

slizhevskyv-semrush:
## Motivation and Context

Type names for `InputFieldProps` on the API page are removed during build because ESLint decided to add `'` since the internal prop is named in the quotes.

## How has this been tested?

It's been tested manually.

Screenshot 2025-07-10 at 12 58 11

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/bulk-textarea/src/components/InputField/InputField.types.ts
+++ b/semcore/bulk-textarea/src/components/InputField/InputField.types.ts
@@ -1,3 +1,5 @@
+/* eslint-disable @stylistic/quote-props */
+
 type PasteProps = {
   /**
    * @default '\n'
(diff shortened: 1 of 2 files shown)


semcore/bulk-textarea/src/components/InputField/InputField.types.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @stylistic/quote-props */
```
ilyabrower: eslint will add quotes without this? but why only here?

slizhevskyv-semrush: since `'aria-describedby'` is in quotes, linter applies this rule to the rest of the properties

ilyabrower: maybe we could write 
```
type InputFieldProps = {
 ....
} & {
  'aria-describedby'?: string;
}
```
?

slizhevskyv-semrush: oh yeah, that will definitely work! :D

slizhevskyv-semrush: updated :)

ilyabrower (approved):

PR #2366, opened 2025-07-29
UIK-3282/focus loss in safari pagination

slizhevskyv-semrush:
## Motivation and Context

In Safari, focus is lost when using `Next`/`Prev` buttons to reach the first or last page. The fix ensures focus remains on the buttons.

## How has this been tested?

I've updated browser test.

## Types of changes

- [x] Nice improve.

## Checklist:

### DIFF
--- a/semcore/pagination/src/Pagination.jsx
+++ b/semcore/pagination/src/Pagination.jsx
@@ -123,11 +123,11 @@ class PaginationRoot extends Component {
       currentPage,
       disabled,
       onClick: () => {
-        const nextPage = currentPage - 1;
+        const followingPage = currentPage - 1;
 
-        this.handlePageChange(nextPage);
+        this.handlePageChange(followingPage);
 
-        if (nextPage <= 1) {
+        if (followingPage <= 1) {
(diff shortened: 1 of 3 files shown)


semcore/pagination/src/Pagination.jsx
```diff
@@ -129,22 +122,38 @@ class PaginationRoot extends Component {
     return {
       currentPage,
       disabled,
-      onClick: () => this.handlePageChange(currentPage - 1),
+      onClick: () => {
+        const nextPage = currentPage - 1;
```
ilyabrower: I understand why, but I don't like `nextPage` here.. What do you think about `followingPage`?) I understand that this is not very good from a language point of view, but maybe it is more correct from a codename point of view?

ilyabrower (approved):

semcore/pagination/src/Pagination.jsx
slizhevskyv-semrush: thanks, updated :)

sheila-semrush (approved):

PR #2453, opened 2025-09-10
[UIK-4155][button] fixed size defaults in types

ilyabrower:
## How has this been tested?
Manually

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/button/src/component/ButtonLink/ButtonLink.type.ts
+++ b/semcore/button/src/component/ButtonLink/ButtonLink.type.ts
@@ -9,7 +9,7 @@ import type {
 
 /**
  *  Button link size
- * @default m
+ * @default 200
  */
 export type ButtonLinkSize = 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800;
 /**
--- a/semcore/button/src/component/ButtonLink/ButtonLink.type.ts
+++ b/semcore/button/src/component/ButtonLink/ButtonLink.type.ts
@@ -9,7 +9,6 @@ import type {
 
 /**
  *  Button link size
- * @default 200
  */
 export type ButtonLinkSize = 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800;
 /**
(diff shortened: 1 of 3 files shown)


slizhevskyv-semrush (approved):

semcore/button/src/component/ButtonLink/ButtonLink.type.ts
```diff
@@ -9,7 +9,7 @@ import type {
 
 /**
  *  Button link size
- * @default m
+ * @default 200
```
sheila-semrush: one caveat, we don't actually have a default size in the component. It's just the CSS that has font-size 200 by default, so it looks like 200, but it doesn't have line-height 200. We should probably either set the default size in `defaultProps` in the component, or remove it here from the prop description

Valeria-Zimnitskaya:
@ilyabrower  could you please fix the size description un AbstractButton.type.ts
Currently we have 
/** Button type. Defined in Button.type or ButtonLink.type */
    size?: S;
Screenshot 2025-09-10 at 13 52 33


Should be size instead of type:
/** Button size. Defined in Button.size or ButtonLink.size */
    size?: S;

semcore/button/src/component/ButtonLink/ButtonLink.type.ts
ilyabrower: I'll remove it here from the prop description

ilyabrower:
> @ilyabrower could you please fix the size description un AbstractButton.type.ts Currently we have /** Button type. Defined in Button.type or ButtonLink.type */ size?: S; Screenshot 2025-09-10 at 13 52 33
> 
> Should be size instead of type: /** Button size. Defined in Button.size or ButtonLink.size */ size?: S;

I changed only first `type`. 2 and 3 - it is correct, because the value defined in types

Valeria-Zimnitskaya (approved):

PR #2459, opened 2025-09-10
[UIK-4149][website] playground - add missing roles for focusable elements

slizhevskyv-semrush:
## Motivation and Context

Focusable elements without roles fail accessibility checks and confuse screen readers

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/playground/components/Code.tsx
+++ b/playground/components/Code.tsx
@@ -103,7 +103,7 @@ function Code({ sourceCode, link }: ICodeProps) {
   if (!html) return;
 
   return (
-    <Box ref={scrollableContainerRef} className={styles.code} aria-label='Source code and links'>
+    <Box ref={scrollableContainerRef} className={styles.code} role='group' aria-label='Source code and actions'>
       <Box className={styles.sourceCodeWrapper}>
         <Box className={styles.sourceCode} dangerouslySetInnerHTML={{ __html: html }}></Box>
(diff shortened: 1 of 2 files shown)


playground/components/Code.tsx
```diff
@@ -103,7 +103,7 @@ function Code({ sourceCode, link }: ICodeProps) {
   if (!html) return;
 
   return (
-    <Box ref={scrollableContainerRef} className={styles.code}>
+    <Box ref={scrollableContainerRef} className={styles.code} aria-label='Source code and links'>
```
sheila-semrush:
```suggestion
    <Box ref={scrollableContainerRef} className={styles.code} aria-label='Source code and actions'>
```

playground/hooks/useFocusableScrollRef.ts
```diff
@@ -15,8 +15,10 @@ function useFocusableScrollRef<T extends HTMLElement>() {
 
       if (hasScroll) {
         node.setAttribute('tabindex', '0');
+        node.setAttribute('role', 'group');
```
sheila-semrush: only it's better to add this in all cases, even when there's no scroll

ilyabrower: I agree with Sheila, just set `role=group` to the `Box` in previous file

sheila-semrush (approved):

PR #2477, opened 2025-09-18
[UIK-4201][chore] fixed unnecessary line breaks

ilyabrower:
## How has this been tested?
Manually

## Types of changes

- [X] Nice improve.

## Checklist:

### DIFF
--- a/playground/components/Playground.tsx
+++ b/playground/components/Playground.tsx
@@ -176,7 +176,6 @@ function Playground<
           sourceCode={reactElementToJSXString(element, {
             showDefaultProps: false,
             filterProps,
-            maxInlineAttributesLineLength: 20,
             ...(JSXDisplayName && { displayName: () => JSXDisplayName }),
           })}
         />


playground/components/Playground.tsx
sheila-semrush, line 1: found 2 places where there's no line wrap even though it should be there:

http://localhost:5173/intergalactic/components/dropdown/dropdown
imagen

http://localhost:5173/intergalactic/components/notice/notice
imagen

Valeria-Zimnitskaya: one more - Screenshot 2025-09-19 at 18 25 36

sheila-semrush: the pagination one is different, I saw it but decided not to mention it. In that case it's just the property value that doesn't fit, which is rare, and this particular value isn't useful anyway. So I'd not do anything with it

Valeria-Zimnitskaya: removed my snapshot)

ilyabrower: This is because `reactElementToJSXString` write only one prop to the same line. 
I can only set maxLenght for the line (and it will move this property to the next, but then you will see the initial issue. 
Let's stay it as is

sheila-semrush: @ilyabrower can't we set maxLength to some bigger value than it was before? 

bigger than here:

imagen

but less than here:

imagen

PR #2487, opened 2025-09-22
[UIK-4227][drag-and-drop] fixed hover styles

ilyabrower:
## How has this been tested?
Manually

## Types of changes

- [X] Nice improve.

## Checklist:

### DIFF
--- a/semcore/drag-and-drop/src/style/drag-and-drop.shadow.css
+++ b/semcore/drag-and-drop/src/style/drag-and-drop.shadow.css
@@ -1,6 +1,5 @@
 SDraggable[draggable] {
   position: relative;
-  cursor: grab;
   -moz-user-select: none;
   -khtml-user-select: none;
   -webkit-user-select: none;
@@ -10,7 +9,7 @@ SDraggable[draggable] {
     cursor: grab;
   }
 
-  &:active {
+  &:active, &[active][isDragging]:hover {
     cursor: grabbing;
   }
 


semcore/drag-and-drop/src/DragAndDrop.tsx
```diff
@@ -121,16 +122,44 @@ class DragAndDropRoot extends Component<AsProps, {}, State> {
       zoneName: zoneName || '',
     });
 
+    let placeholder: HTMLDivElement | null = null;
+
+    if (event.type === 'dragstart') {
+      placeholder = document.createElement('div');
+      placeholder.style.backgroundColor = 'var(--intergalactic-bg-secondary-neutral, #f4f5f9)';
```
sheila-semrush:
```suggestion
      placeholder.style.backgroundColor = 'var(--intergalactic-bg-primary-neutral-hover, #f4f5f9)';
```

semcore/drag-and-drop/src/style/drag-and-drop.shadow.css
sheila-semrush, line 1: we've lost the grab cursor :)
it should be `grab` while hovering over the draggable object, and also `grabbing` while dragging (the last one wasn't implemented before, but would be very nice if we could do it now)
btw, in firefox there's no automatic cursor change while dragging, so the cursor currently is always `pointer`

ilyabrower: As I remember, we don't need it over all draggable element, only over the "points". But in options with checkboxes we can't render them, because of they are rendering in `::before` element as a part of checkboxes.. so...

ilyabrower: fixed

semcore/drag-and-drop/src/DragAndDrop.tsx
sheila-semrush, line 1: @ilyabrower found one small bug: in Chrome, when dropping an item, the cursor remains "grabbed", and the item that's now in the old position of the dragged item becomes hovered. It's present on prod as well. So you can decide whether we should fix it now or leave for the future

sheila-semrush: video:
https://github.com/user-attachments/assets/7d5541c6-ca41-4ec9-aae0-de5aee970f96

ilyabrower: yes, let's fix it in some another bug

PR #2491, opened 2025-09-22
[UIK-4251][website] fixed role for basic example with scroll area

ilyabrower:
## How has this been tested?
Manually

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/base-components/src/components/scroll-area/ScrollArea.tsx
sheila-semrush, line 1: @ilyabrower maybe let's not do this change in the component? because the role can vary greatly depending on the specific usage. We can add the group role just in the example

PR #2502, opened 2025-09-30
[UIK-4267][data-table] accordion + checkbox

ilyabrower:
## How has this been tested?
Manually + I added a story

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/data-table/CHANGELOG.md
```diff
@@ -2,12 +2,16 @@
...
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
-## [16.4.0] - 2025-09-20
+## [16.4.0] - 2025-10-03
 
 ### Added
 
 - Limited mode support.
 
+### Fixed
```
sheila-semrush: don't you want to present this as "added"? I think this update deserves more attention :)

ilyabrower: changed

ilyabrower:
> @ilyabrower everything looks good here
> 
> Found one small problem, but it existed before this change, so you can decide whether to fix it now or later:
> 
> With sideIndents='wide' or variant='card', left/right paddings in the accordion's nested cell should also increase imagen

Fixed in this PR

semcore/data-table/src/components/Body/style.shadow.css
sheila-semrush, line 332: only in CollapseRow the attribute is "aria-level", so this doesn't apply (the resulting padding is 2px bigger than should be)

ilyabrower: thank you, fixed

sheila-semrush (approved):

Valeria-Zimnitskaya (approved):

PR #2504, opened 2025-09-30
[UIK-4166][tooltip] fixed bottom scrollbar appearance in overflowed modals

slizhevskyv-semrush:
## Motivation and Context

Tooltip triggers a bottom scrollbar in overflowed modals because the initial check assumes the scroll container is the `body`. Added logic to handle this by detecting the first parent scroll container.

## How has this been tested?

I've added a story and a browser test.

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/core/src/utils/use/useScrollBarWidth.ts
```diff
@@ -1,19 +1,46 @@
...
+  return findScrollableParentFor(element.parentElement);
+}
+
+function getParentScrollableElement(ref: React.RefObject<HTMLElement>) {
+  if (!ref || !ref.current) return null;
+
+  return findScrollableParentFor(ref.current);
+}
+
+export function useScrollBarWidth(ref: React.RefObject<HTMLElement>, callback: (width: number) => void = () => {}, vertical = true) {
```
ilyabrower: I'm not sure, that we need default value for the `callback`...

semcore/core/src/utils/use/useScrollBarWidth.ts
```diff
@@ -1,19 +1,46 @@
...
     const calculateScrollBar = () => {
       if (!window.visualViewport) return;
 
-      if (!vertical) {
-        setScrollBarWidth(window.innerHeight - window.visualViewport.height);
-        return;
+      let width = vertical ? window.innerWidth - window.visualViewport.width : window.innerHeight - window.visualViewport.height;
+
+      if (!width) {
+        const scrollableParenElement = getParentScrollableElement(ref);
```
ilyabrower: maybe we should check `ref.current` on the line above and use `findScrollableParentFor` here?

semcore/core/src/utils/use/useScrollBarWidth.ts
slizhevskyv-semrush: agree :)

semcore/core/src/utils/use/useScrollBarWidth.ts
slizhevskyv-semrush: yeas, good point!

PR #2550, opened 2025-10-24
[UIK-3090][dropdown] prevented default behaviour for arrow up & down

slizhevskyv-semrush:
## Motivation and Context

Prevented default behaviour for `Arrow.Up` & `Arrow.Down` buttons for `Dropdown` and it's ancestors.

## How has this been tested?

Manually

## Types of changes

- [x] Nice improve.

## Checklist:


semcore/dropdown/src/AbstractDropdown.tsx
```diff
@@ -271,6 +271,8 @@ export abstract class AbstractDropdown extends Component<AbstractDDProps, {}, {}
     ) {
       if (this.asProps.visible !== true) {
         if (['ArrowDown', 'ArrowUp'].includes(e.key)) {
+          e.preventDefault();
```
sheila-semrush, line 345: on one hand, it seems to work as expected (scroll prevented with dropdown menu, but still happens with the common dropdown), but is it right to add this logic on the level of Dropdown? 
considering only the current task scope, page scroll should only be prevented for DropdownMenu

BUT, after looking at other examples, ideally we should also prevent page scroll AND show popper for Dropdown.Trigger with `role=combobox`, regardless if there's menu or not (which doesn't happen currently)
one case where dropdown (not menu) trigger should also open on Up/Down arrow is this (because trigger has `combobox` role): http://localhost:5173/intergalactic/filter-group/advanced-filters/advanced-filters-code

I'm mentioning this just in case, we can separate the latter into another task, but maybe it can be useful in this one

slizhevskyv-semrush: decided to create a separate task for this behaviour

semcore/dropdown/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 CHANGELOG.md standards are inspired by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [16.1.3] - 2025-11-14
+
+### Changed
+
+- Canceled default behaviour for `Arrow Up` `Arrow Down` keys.
```
sheila-semrush:
```suggestion
- Canceled default behavior for `Arrow Up` `Arrow Down` keys.
```
let's listen to the doc linter)

PR #2579, opened 2025-11-12
[docs] updated product head docs

sheila-semrush:
## Motivation and Context

Removed some outdated info, updated pics, fixed typos.

## Types of changes

- [x] Nice improve.

## Checklist:


j-mnizhek (dismissed):
👍

Just a small comment
I'm a little bit worried that we have illustrations with hardcoded values for paddings and margins, since they might change in the future. Was thinking maybe we should use token names there, if there is no good option to remove these illustrations

sheila-semrush:
> 👍
> 
> Just a small comment I'm a little bit worried that we have illustrations with hardcoded values for paddings and margins, since they might change in the future. Was thinking maybe we should use token names there, if there is no good option to remove these illustrations

true, although I think it's more likely that we might change the number of spacing units in some paddings/margins (like spacing-2x to spacing-3x and so on), than change the value of the base spacing unit itself. And in this case we will still have to mass update illustrations. Maybe if we introduced some higher level spacing tokens, like spacing-m/l/xl etc, this problem could be solved at least to some extent.

PR #2606, opened 2025-11-24
[UIK-4551][website] update Export dropdown and FeedbackYesNo docs

j-mnizhek:
## Motivation and Context

- Removed unnecessary info from the doc about our Export dropdown pattern.
- Renamed it to make the pattern clear.
- Removed unnecessary illustration from FeedbackYesNo doc.

## Types of changes

- [x] Nice improve.


j-mnizhek:
> Everything looks good! I just have one small question for http://localhost:5173/intergalactic/patterns/feedback-yes-no/feedback-yes-no page Should the “Close” button be without quotes? (I'm asking because other buttons have quotes.)
> 
> BTW, we already have this in prod, but if this is unexpected

Thank you for the question! ✨ It's okay for Close button to live without quotes because it's its function/purpose/whatever: to close, and we can omit quotes in this case. "Ask me later" button has quotes because in fact it's the name of the button.

PR #2630, opened 2025-12-12
[UIK-4606][notice-global] deprecate component

slizhevskyv-semrush:
#### Deprecated

- Package is deprecated.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/notice-global/src/index.d.ts
+++ b/semcore/notice-global/src/index.d.ts
@@ -29,11 +29,11 @@ export type NoticeGlobalProps = FadeInOutProps & {
   locale?: string;
 };
 
-/** @deprecated since v.17 */
+/** @deprecated since v.16 -> will be removed in v.17 */
 declare const NoticeGlobal: Intergalactic.Component<'div', NoticeGlobalProps> & {
-  /** @deprecated since v.17 */
+  /** @deprecated since v.16 -> will be removed in v.17 */
   Content: typeof Flex;
-  /** @deprecated since v.17 */
+  /** @deprecated since v.16 -> will be removed in v.17 */
   CloseIcon: Intergalactic.Component<'div', IconProps>;
 };
 
--- a/semcore/notice-global/README.md
+++ b/semcore/notice-global/README.md
@@ -2,7 +2,7 @@
 
 # @semcore/notice-global (DEPRECATED)
 
-## This package is deprecated and will be deleted in the future releases.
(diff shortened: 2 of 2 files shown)


semcore/notice-global/src/index.d.ts
```diff
@@ -29,8 +29,11 @@ export type NoticeGlobalProps = FadeInOutProps & {
   locale?: string;
 };
 
+/** @deprecated since v.16 */
```
ilyabrower: v17, i think?

slizhevskyv-semrush: yes, thanks!

ilyabrower: and two times below =)

ilyabrower (dismissed):

semcore/notice-global/src/index.d.ts
```diff
@@ -29,8 +29,11 @@ export type NoticeGlobalProps = FadeInOutProps & {
   locale?: string;
 };
 
+/** @deprecated since v.17 */
```
sheila-semrush: probably also need to add this info to package's README

semcore/notice-global/src/index.d.ts
```diff
@@ -29,8 +29,11 @@ export type NoticeGlobalProps = FadeInOutProps & {
...
 declare const NoticeGlobal: Intergalactic.Component<'div', NoticeGlobalProps> & {
+  /** @deprecated since v.17 */
   Content: typeof Flex;
+  /** @deprecated since v.17 */
```
sheila-semrush: is "deprecated since v.17" correct if we are going to remove the package in v.17? or we're not?
maybe it will be more accurate to write `deprecated, will be removed in v.17`

sheila-semrush (dismissed):

semcore/notice-global/src/index.d.ts
slizhevskyv-semrush: sure, good point! Thanks!

semcore/notice-global/src/index.d.ts
slizhevskyv-semrush: I expected that it would be only deprecated in v17 and removed in v.18 or later. That was my thought.

slizhevskyv-semrush: @ilyabrower do we plan to remove it in 18+?

PR #2656, opened 2026-01-13
[UIK-3535][core] update values for lh variables to round them in browsers

j-mnizhek:
#### Changed

- Updated the values ​​of variables in the `lh` group: unitless values ​​are now used instead of percentages.

## Motivation and Context

- In Chrome, Mozilla, and Safari, the new values ​​are rounded uniformly, instead of as percentages.
- Updated test story for Typography, added explicitly `size` prop for 16px text. Without this, some default styles were applied for the 16px text. 

## How has this been tested?

Manually, with storybook and browsers.

## Types of changes

- [x] Nice improve.

## Checklist:


j-mnizhek:
Added these changes to https://github.com/semrush/intergalactic/pull/3130

PR #2657, opened 2026-01-14
[UIK-2389][fullscreen-modal] update Header styles

j-mnizhek:
#### Changed

- Header styles, for the correct text behaviour and design ofc ✨

## Motivation and Context

The previous design decision to place the description text on the same line with the title isn’t very good, both visually and technically, so I:
- updated the layout to a column
- added a `<p>` tag to the description
- added a test story to check the behavior of a long title and description
- updated documentation accordingly

## How has this been tested?

Manually, with stories and different browsers

## Screenshots

Current
image

New one
image

## Types of changes

- [x] Nice improve.

## Checklist:

### DIFF
--- a/semcore/notice-bubble/__tests__/index.test.tsx
+++ b/semcore/notice-bubble/__tests__/index.test.tsx
@@ -83,7 +83,7 @@ describe('NoticeBubbleContainer', () => {
 
   test('Verify notices render in custom container via ref when containerNode is RefObject', async () => {
     const TestComponent = () => {
-      const containerRef = React.useRef<HTMLDivElement>(null);
+      const [containerNode, setContainerNode] = React.useState<HTMLElement | null>(null);
       const [manager] = React.useState(() => new NoticeBubbleManager());
 
       React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


semcore/fullscreen-modal/src/style/fullscreen-modal.shadow.css
```diff
@@ -55,9 +55,9 @@ SBackText {
 
 STitle,
 SDescription {
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
+  overflow-wrap: break-word;
+  word-break: break-word;
+  white-space: normal;
```
ilyabrower: not sure that we need this default

j-mnizhek: Agree! I'd leave only
`overflow-wrap: break-word;`
because we cannot predict the length of the words in different languages, so they need to be broken to the next line instead of being cropped and hidden. Though it's a really rare case, but left this solution.

PR #2678, opened 2026-01-22
[UIK-3737][website] added ancors to tables with props on api-pages

ilyabrower:
## Types of changes

- [X] Nice improve.

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/semcore/feedback-form/src/component/feedback-rating/FeedbackRating.tsx
+++ b/semcore/feedback-form/src/component/feedback-rating/FeedbackRating.tsx
@@ -396,7 +396,7 @@ function Header(props: any) {
   const { styles } = props;
   const SHeader = Root;
   return sstyled(styles)(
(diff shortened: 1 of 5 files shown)


website/docs/.vitepress/buildHooks.ts
```diff
@@ -210,6 +210,11 @@ const transformPageData: UserConfig<DefaultTheme.Config>['transformPageData'] =
 
   if (!folder || !parentName || !pageName) return;
 
+  // Enable h3 in outline for API pages
```
sheila-semrush: we don't want them in the outline, actually (we don't have h3 links in the outline anywhere else)

sheila-semrush: (links to these headings will work without them being in the outline)

PR #2679, opened 2026-01-26
[website] removed h3 from outline, fixed some API pages

sheila-semrush:
## Motivation and Context

I removed H3 headings from the outline on API pages because it was redundant in the majority of the cases. H3 headings almost always duplicate parent headings:

imagen

Also fixed some specific API pages: fixed structure, added missing types.

We could get rid of manually added H2 headings and replace auto generated type name headings with H2 level, but there are some concerns:

1. There's often content before the H3 heading (some short text about the component). If we want to get rid of manual H2 headings, we need to design a way to feed this content into the TypesView VitePress component.
1. Type names are often harder to read or not very descriptive (`InputTags.Value` vs `InputTagsValueProps`). I'm also not sure it's beneficial for anyone to know the type names.
2. There are pages where there are deprecated types, and it's better to group/label them. Example: [DropdownMenu](http://localhost:5173/intergalactic/components/dropdown-menu/dropdown-menu-api).
3. There are 1-2 pages where there's too many types, and it's better to group them. Main case is [d3 chart](http://localhost:5173/intergalactic/data-display/d3-chart/d3-chart-api).

In conclusion, I think it would be easier to stay with only the manually edited H2 headings in the outline for now.

I have some ideas how we could improve API pages structure, but it would require more work, so it's better to do it in a separate task.

## Types of changes

- [x] Nice improve.

## Checklist:

### DIFF
--- a/website/docs/components/bulk-textarea/bulk-textarea-api.md
+++ b/website/docs/components/bulk-textarea/bulk-textarea-api.md
@@ -7,13 +7,46 @@ tabs: Design('bulk-textarea'), A11y('bulk-textarea-a11y'), API('bulk-textarea-ap
 
 ```jsx
 import BulkTextarea from '@semcore/ui/bulk-textarea';
-<BulkTextarea>
-  <BulkTextarea.InputField/>
-</BulkTextarea>;
+<BulkTextarea />
+```
+
+<TypesView type="BulkTextareaProps" :types={...types} />
+
+## BulkTextarea.InputField
+
+```jsx
+import BulkTextarea from '@semcore/ui/bulk-textarea';
+<BulkTextarea.InputField />
 ```
 
 <TypesView type="InputFieldProps" :types={...types} />
 
-<TypesView type="BulkTextareaProps" :types={...types} />
+## BulkTextarea.ErrorsNavigation
+
+```jsx
+import BulkTextarea from '@semcore/ui/bulk-textarea';
+<BulkTextarea.ErrorsNavigation />
+```
+
+<TypesView type="ErrorsNavigationProps" :types={...types} />
+
+## BulkTextarea.ClearAll
+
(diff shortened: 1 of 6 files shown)


website/docs/components/checkbox/checkbox-api.md
```diff
@@ -41,8 +40,6 @@ import Checkbox from '@semcore/ui/checkbox';
 <Checkbox.Value.CheckMark />;
 ```
 
-<TypesView type="CheckboxValueCheckMarkProps" :types={...types} />
```
ilyabrower, line 44: Not sure this is good, that you removed that and previous types

website/docs/components/bulk-textarea/bulk-textarea-api.md
```diff
@@ -7,13 +7,46 @@ tabs: Design('bulk-textarea'), A11y('bulk-textarea-a11y'), API('bulk-textarea-ap
 
 ```jsx
 import BulkTextarea from '@semcore/ui/bulk-textarea';
-<BulkTextarea>
-  <BulkTextarea.InputField/>
-</BulkTextarea>;
```
ilyabrower, line 12: You can't use it with just `<BulkTextarea />`, so, the only one example you need here is 
```
<BulkTextarea>
  <BulkTextarea.InputField/>
</BulkTextarea>;
```

website/docs/data-display/mini-chart/mini-chart-api.md
```diff
@@ -13,25 +13,31 @@ import MiniChart from '@semcore/ui/mini-chart';
 
 <TypesView type="ScoreLineGaugeProps" :types={...types} />
 ```jsx
+import MiniChart from '@semcore/ui/mini-chart';
```
ilyabrower, line 16: I'm not sure about this and all imports below... We show how to import it in the first example, should we just copy-paste the same code?

PR #2684, opened 2026-01-28
[UIK-4704][d3-chart] added showPercentValueInTooltip prop for Area/Bar/Cigarette/Histogram/Line charts + updated types

slizhevskyv-semrush:
#### Added

- `showPercentValueInTooltip` prop for Cigarette chart.

#### Fixed
- Fixed types by removing unsupported `showTotalInTooltip` prop from certain chart components.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/drag-and-drop/__tests__/drag-and-drop.browser-test.tsx
+++ b/semcore/drag-and-drop/__tests__/drag-and-drop.browser-test.tsx
@@ -181,31 +181,26 @@ test.describe(`${TAG.FUNCTIONAL} `, () => {
       await page.keyboard.press('Tab');
       await expect(locators.dropZone(page).first()).toBeFocused();
       await page.keyboard.press('Tab');
+      await page.waitForTimeout(100); // added timeout because of slow animation of this component
       await expect(locators.draggable(page).first()).toBeFocused();
       await test.step('Verify graggable items can be moved to the drop zone', async () => {
         await page.keyboard.press('Space');
-        await expect.poll(
-          async () => {
-            return await locators.draggable(page).first().getAttribute('class');
-          },
-          {
-            timeout: 200,
-          },
-        ).toMatch(/keyboardDragging/);
+        await page.waitForTimeout(100);
+
+        await expect(locators.draggable(page).first()).toHaveClass(/keyboardDragging/, { timeout: 200 });
         await expect(page.getByRole('alert')).toHaveCount(1);
         await page.keyboard.press('ArrowLeft');
+        await page.waitForTimeout(100);
+
         await page.keyboard.press('Space');
+        await page.waitForTimeout(100);
+
         await expect(locators.draggable(page).first()).toBeFocused();
 
-        await expect.poll(
-          async () => {
-            return await locators.draggable(page).first().getAttribute('class');
-          },
-          {
-            timeout: 200,
-          },
-        ).not.toMatch(/keyboardDragging/);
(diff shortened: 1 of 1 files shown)


semcore/d3-chart/src/component/Chart/AbstractChart.tsx
```diff
@@ -266,6 +267,25 @@ export abstract class AbstractChart<
...
+
+    const value = data[key];
+
+    if (typeof value === 'number') {
+      // just to prevent NaN and Infinity as procent.
+      if (value === 0 || total === 0) {
+        return `0%`;
+      }
+
+      const procent = Math.round((100 * value) / total);
```
ilyabrower:
```suggestion
      const percent = Math.round((100 * value) / total);
```

semcore/d3-chart/src/component/Chart/AbstractChart.tsx
```diff
@@ -266,6 +267,25 @@ export abstract class AbstractChart<
...
+
+      return `${procent}%`;
```
ilyabrower:
```suggestion
      return `${percent}%`;
```

playground/entries/Chart/Cigarette.tsx
```diff
@@ -0,0 +1,81 @@
...
+  Dogs: 1344,
+  Capybaras: 6135,
+  Hamsters: 1456,
+  Birds: 1823,
+};
+
+function getJSX(props: CigaretteChartJSXProps) {
+  const { ...legendProps } = props.legendProps ?? {};
+  return (
+    <Chart.Cigarette
```
sheila-semrush, line 34: @slizhevskyv-semrush I have a concern about the display name, isn't "Cigarette.Bar" invalid here?
image

playground/entries/Chart/Cigarette.tsx
```diff
@@ -0,0 +1,81 @@
...
+      data={data}
+      plotWidth={props.cigaretteProps.layout === 'horizontal' ? 300 : 44}
+      plotHeight={props.cigaretteProps.layout === 'horizontal' ? 28 : 200}
+      aria-label='Cigarette chart'
+      tooltipViewType={props.cigaretteProps.tooltipViewType}
+      invertAxis={props.cigaretteProps.layout !== 'vertical'}
```
sheila-semrush, line 40: @slizhevskyv-semrush there's also an issue with `invertAxis={false}`, I guess we hide it, but in this case the default value is `true` and we should show it explicitly when it's `false` 

Also would be nice if `undefined` hid properties, now it's printed in the snippet as is. In this case I'm expecting:

-  `layout="horizontal"` => `invertAxis={undefined}` (hidden from snipped, default value), 
- `layout="vertical"` => `invertAxis={false}` (printed in snippet)

In general, I wouldn't hide `false`, but instead would hide `undefined`.
image

sheila-semrush: created a sub task for it UIK-4807

playground/entries/Chart/Cigarette.tsx
ilyabrower: fixed

playground/entries/Chart/Cigarette.tsx
ilyabrower: fixed

sheila-semrush (dismissed):

PR #2690, opened 2026-01-30
UIK-3093/feedback-rating-a11y-issue

slizhevskyv-semrush:
#### Fixed

- Added conditional logic for displaying `aria-label` based on `aria-labelledby` prop.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/feedback-form/src/style/feedback-rating.shadow.css
+++ b/semcore/feedback-form/src/style/feedback-rating.shadow.css
@@ -11,7 +11,7 @@ SFeedbackRating {
     font-size: var(--intergalactic-fs-300, 16px);
     line-height: var(--intergalactic-lh-300, 150%);
     font-weight: var(--intergalactic-bold, 700);
-    margin: var(--intergalactic-spacing-4x) 0;
+    margin: var(--intergalactic-spacing-4x, 16px) 0;
     text-align: center;
   }
 }


semcore/feedback-form/src/component/feedback-rating/FeedbackRating.tsx
```diff
@@ -396,7 +396,7 @@ function Header(props: any) {
   const { styles } = props;
   const SHeader = Root;
   return sstyled(styles)(
-    <SHeader render={Text} size={300} tag='h2' mb={4} mt={4} textAlign='center' />,
+    <SHeader render={Modal.Title} tag='h2' mb={4} mt={4} textAlign='center' />,
```
ilyabrower: I think, Modal.Title should be Text under the hood, so we can use size=300

slizhevskyv-semrush: Yeah, you are right, it should work, but `Title` inside modal set `font-size` in styles. That is why it doesn't work :)

semcore/feedback-form/src/component/feedback-rating/FeedbackRating.tsx
```diff
@@ -396,7 +396,7 @@ function Header(props: any) {
   const { styles } = props;
   const SHeader = Root;
   return sstyled(styles)(
-    <SHeader render={Text} size={300} tag='h2' mb={4} mt={4} textAlign='center' />,
+    <SHeader render={Modal.Title} mb={4} mt={4} textAlign='center' />,
```
ilyabrower: you have `margin: 0` in css, so, let's remove `mb` and `mt`.

slizhevskyv-semrush: I need `margin: 0` to reset `margin` behaviour from `Modal.Title`. `mb={4}` & `mt={4}` applied as inline styles for element.

Let's made this way -> i'll remove those annotations from `tsx` file and move it to css. To make it defined at one place :)

ilyabrower (approved):

PR #2714, opened 2026-02-12
[UIK-4742][d3-chart] add minimalBarWidth property for Cigarette chart

slizhevskyv-semrush:
#### Added

- `minimalBarWidth` for `Cigarette` chart.

## Motivation and Context

Added `minimalBarWidth` property for `Cigarette` chart.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/d3-chart/__tests__/cigarette-chart.browser-test.tsx
+++ b/semcore/d3-chart/__tests__/cigarette-chart.browser-test.tsx
@@ -201,39 +201,133 @@ test.describe(`${TAG.FUNCTIONAL}`, () => {
         await expect(item).toHaveAttribute('direction', 'horizontal');
       }
     });
+  });
 
-    await test.step('Verify minimal bar width for small values', async () => {
-      const data = {
-        Cats: 3524,
-        Dogs: 1,
-        Capybaras: 6135,
-        Hamsters: 1,
-        Birds: 1823,
-      };
-      const minimalBarWidth = 8;
-      const smallItemKeys = ['Dogs', 'Hamsters'];
-
+  test('Verify minimalBarWidth behavior', {
+    tag: [TAG.PRIORITY_HIGH, '@cigarette-chart', '@d3-chart'],
+  }, async ({ page }) => {
+    await test.step('Verify tiny value invisible at minimalBarWidth=0', async () => {
       await loadPage(
         page,
         'stories/components/d3-chart/tests/examples/cigarette-chart/basic-usage.tsx',
         'en',
-        {
-          data,
-          minimalBarWidth,
-        },
+        { data: { Large: 10000, Medium: 5000, Tiny: 1 }, minimalBarWidth: 0, duration: 0 },
       );
 
       await locators.plot(page).first().waitFor({ state: 'visible' });
+      await page.waitForTimeout(300);
+      await expect(locators.barItem(page)).toHaveCount(3);
+
       const items = await locators.barItem(page).all();
-
       for (const item of items) {
-        const itemTextContent = await item.textContent();
-
-        if (!itemTextContent) continue;
-
-        if (smallItemKeys.includes(itemTextContent)) {
-          await expect(item).toHaveAttribute('width', `${minimalBarWidth}`);
+        const text = await item.textContent();
+        if (text?.trim() === 'Tiny') {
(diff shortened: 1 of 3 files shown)


semcore/d3-chart/src/component/Chart/AbstractChart.tsx
```diff
@@ -63,9 +59,15 @@ export abstract class AbstractChart<
     this.resolveColor = this.resolveColor.bind(this);
     this.tooltipValueFormatter = this.tooltipValueFormatter.bind(this);
     this.handleWithTrendChange = this.handleWithTrendChange.bind(this);
+
+    this.state = {
+      dataDefinitions: this.getDefaultDataDefinitions(),
+      highlightedLine: -1,
+      withTrend: false,
+    } as State;
```
ilyabrower, line 67: let's remove `as`

semcore/d3-chart/src/component/Chart/CigaretteChart.tsx
```diff
@@ -52,10 +63,38 @@ class CigaretteChartComponent extends AbstractChart<
...
+    return dataDefinitions.filter(({ checked }) => checked);
+  }
+
+  private get nonNullActiveDataDefinitions() {
+    const { data } = this.asProps;
+
+    return this.activeDataDefinitions.filter(({ id }) => {
+      const itemValue = data[id];
+
+      return typeof itemValue === 'number' && itemValue > 0;
```
ilyabrower, line 96: technically, this is not `nonNull`, but something like `positive` or `present` values... just to think about it

semcore/d3-chart/src/component/Chart/CigaretteChart.tsx
```diff
@@ -78,16 +117,109 @@ class CigaretteChartComponent extends AbstractChart<
     };
   }
 
+  private onPlotMouseMove = trottle((event: React.MouseEvent<HTMLElement>) => {
+    if (!this.plotRef.current) return;
+
+    const [pX, pY] = eventToPoint(event, this.plotRef.current);
+
+    this.setState((prevState) => ({ ...prevState, pX, pY }));
```
ilyabrower: do we really need function? why not `this.setState({ pX, pY });`?
I'm just surprised =)

semcore/d3-chart/src/component/Chart/CigaretteChart.tsx
```diff
@@ -311,42 +455,31 @@ class CigaretteChartComponent extends AbstractChart<
...
+          const lastAddedDomain = domain[domain.length - 1];
+          domain.push(lastAddedDomain + visualWidth + DEFAULT_GAP);
+        }
+      }
 
-    this.selectedData.forEach((value) => {
-      max = max + value;
-    });
+      return acc;
+    }, { range: [], domain: [] } as { range: Array<CigaretteChartDataKey>; domain: Array<number> });
```
ilyabrower: let's add `type ScaleThreshold = { range: Array<CigaretteChartDataKey>; domain: Array<number> }` and use it as `cigaretteItems.reduce<ScaleThreshold>()`

semcore/d3-chart/src/component/Chart/AbstractChart.tsx
slizhevskyv-semrush: it's needed since `ts` initializes state as literal object with type `{ highlightedLine: -1,...` rather than `{ highlightedLine: number,...}`

semcore/d3-chart/src/component/Chart/CigaretteChart.tsx
slizhevskyv-semrush: yeah, missed that moment, thanks!

semcore/d3-chart/src/component/Chart/AbstractChart.tsx
ilyabrower: `this.state: State = {...};` ?

because with `as` solution, you can set `highlightedLine: 'qwe'` for example =)

slizhevskyv-semrush: you cannot define type of the variable inside constructor like this :)
yeah, you are right, it's a bit unsafe but looks like it's the only possible solution here :D

PR #2730, opened 2026-02-19
UIK-4841/button-link-underline-v-17

slizhevskyv-semrush:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/link/__tests__/link.browser-test.tsx
+++ b/semcore/link/__tests__/link.browser-test.tsx
@@ -106,6 +106,7 @@ test.describe(` ${TAG.VISUAL}`, () => {
           size, active: true, ...addonVars, text: longText, ...ellipsisVars, ...extraVars,
         });
         const activeClip = await getTextClip(page);
+        await page.waitForTimeout(200); // Finish for ellipsis apply
 
         await test.step('Verify underlined state + hover hint', async () => {
           await locators.link(page).first().hover();
@@ -146,6 +147,7 @@ test.describe(` ${TAG.VISUAL}`, () => {
       size: 400, noWrap: true, ellipsis: false, w: 200,
     });
     const clip = await getTextClip(page);
+    await page.waitForTimeout(200); // Finish for ellipsis apply
 
     await test.step('Verify focus on first link', async () => {
       await page.keyboard.press('Tab');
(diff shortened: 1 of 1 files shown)


semcore/button/src/component/ButtonLink/buttonLink.shadow.css
```diff
@@ -41,8 +36,15 @@ SButton {
...
 
   SText {
-    box-shadow: 0 1px 0 0 transparent;
-    transition: box-shadow 0.15s ease-in-out;
+    text-decoration: underline;
+    text-decoration-style: solid;
+    text-decoration-thickness: 1px;
+    text-decoration-color: transparent;
+    text-underline-position: under;
+    text-underline-offset: 1px;
```
ilyabrower: maybe offset should be 2px?

slizhevskyv-semrush: idk, it could be tweaked :) 
probably some input is required from @j-mnizhek or @sheila-semrush

sheila-semrush: you can check whether the offset looks the same as in Link on hover
(also offset should be the same in primary buttonlink on hover and in secondary buttonlink)
we will look at it on design review and change the value if needed, but it will be nice if you make sure everything looks consistent at least in some degree

sheila-semrush: (but 2px sounds legit)

slizhevskyv-semrush: yeah, I tried to make it unified as it possible, but the `text-decoration` depends on some `font` variables that could affect at which point offset the `text-decoration` is painted. You could check how `ButtonLink` looks like for different sizes in comparison to `box-shadow`. 

I just left `1px` since it's the most closest to the old implementation value.

slizhevskyv-semrush: In Figma, this component has slightly different css properties. At least it has `text-underline-position: from-font;` which work fine for lines like `ButtonLink` but doesn't work fine for lines like this `Buttonyes`. Where the `text-decoration` will be stuck to the baseline of letter `y`.

PR #2766, opened 2026-03-03
[UIK-4866] Small link/button-link underline style fix

sheila-semrush:
## Motivation and Context

- simplified underlining styles (moved underline slightly closer to text)
- removed `bottom-padding` — I think it's not needed now, but I'm not sure I remember correctly for which case it was added
- fixed `text-decoration-thickness` not applying in secondary button link
- added explicit size to some link examples (underlining is cropped without set line-height)

## How has this been tested?

Manually.
Not uploading snapshots here because they haven't been updated in the parent branch, so they would contain some other changes apart from these.

## Types of changes

- [ ] Nice improve.

## Checklist:


semcore/button/src/component/ButtonLink/buttonLink.shadow.css
```diff
@@ -36,16 +36,13 @@ SButton {
   }
 
   SText {
-    text-decoration: underline;
+    text-decoration-line: underline;
```
ilyabrower: why do you use `text-decoration-line` here and without `-line` in Link?

ilyabrower:
> * removed `bottom-padding` — I think it's not needed now, but I'm not sure I remember correctly for which case it was added

Did you check it with `ellipsis` settings? It will enable `overflow: hidden`

semcore/button/src/component/ButtonLink/buttonLink.shadow.css
sheila-semrush: here text-decoration in the shorthand form caused `text-decoration-thickness` to not apply, in link there's no such problem, but we can do the same there for consistency. It's probably better to not mix shorthands and single attributes anyway

sheila-semrush: actually no, I was wrong, the problem with thickness was only in the secondary use, and what fixed it was removing `text-decoration: underline` from `SButton[use='secondary'] SText` due to order of rules. So I will roll back this change

stories/components/link/tests/examples/basic_usage.tsx
```diff
@@ -69,6 +70,8 @@ const Demo = (props: BasicLinkProps) => {
         >
           <Link.Text w={w} ellipsis>{text} ellipsis</Link.Text>
         </Link>
+      </Text>
+      <Text tag='div' mt={3} size={size}>
```
ilyabrower: maybe it's better to remove this to render both links on the same line?

ilyabrower (approved):

stories/components/link/tests/examples/basic_usage.tsx
sheila-semrush: fixed

PR #2802, opened 2026-03-17
[UIK-4947][d3-chart] add percent formatter for Cigarette chart

slizhevskyv-semrush:
#### Added

- `percentFormatter` for Cigarette.

## Motivation and Context

Added `percentFormatter` prop for `Chart.Cigarette`.

## How has this been tested?
 I've added unit test.

## Types of changes

- [ ] Nice improve.

## Checklist:

### DIFF
--- a/semcore/d3-chart/__tests__/index.test.tsx
+++ b/semcore/d3-chart/__tests__/index.test.tsx
@@ -763,6 +763,8 @@ describe('Chart.Cigarette', () => {
   beforeEach(cleanup);
 
   test.concurrent('should call percentFormatter and return correct formatted percent', async () => {
+    vi.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => (cb as any)());
+
     const percentFormatter = vi.fn((value: number) => value.toFixed(2));
 
     const { getByLabelText } = render(
@@ -794,6 +796,8 @@ describe('Chart.Cigarette', () => {
     expect(percentFormatter).toHaveNthReturnedWith(1, '23.81');
     expect(percentFormatter).toHaveNthReturnedWith(2, '52.38');
     expect(percentFormatter).toHaveNthReturnedWith(3, '23.81');
+
+    (window.requestAnimationFrame as any).mockRestore();
   });
 });
 
--- a/semcore/d3-chart/__tests__/index.test.tsx
+++ b/semcore/d3-chart/__tests__/index.test.tsx
@@ -759,6 +759,46 @@ describe('Chart.Venn', () => {
   });
 });
 
+describe('Chart.Cigarette', () => {
+  beforeEach(cleanup);
+
+  test.concurrent('should call percentFormatter and return correct formatted percent', async () => {
+    vi.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => (cb as any)());
+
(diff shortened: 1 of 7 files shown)


semcore/d3-chart/__tests__/index.test.tsx
```diff
@@ -759,6 +759,46 @@ describe('Chart.Venn', () => {
   });
 });
 
+describe('Chart.Cigarette', () => {
+  beforeEach(cleanup);
+
+  test.concurrent('should call percentFormatter and return correct formatted percent', async () => {
+    vi.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => (cb as any)());
```
ilyabrower, line 766: if we have to use `requestAnimationFrame`, let's move it to global mocks

semcore/d3-chart/src/component/Chart/CigaretteChart.tsx
```diff
@@ -369,6 +369,27 @@ class CigaretteChartComponent extends AbstractChart<
...
+  protected percentValue(data: ObjectData, key: string): string {
+    const { percentFormatter } = this.asProps;
+
+    const total = this.totalValue();
+
+    const value = data[key];
+
+    if (typeof value === 'number' && total !== 0) {
+      const rawPercent = (100 * value) / total;
+      const formattedPercent = percentFormatter ? percentFormatter(rawPercent) : Math.round(rawPercent);
```
ilyabrower, line 381: i think, we should add this prop to AbstractChart and add this logic in its percentValue method...

semcore/d3-chart/src/component/Chart/CigaretteChart.type.ts
```diff
@@ -18,6 +18,8 @@ export type CigaretteChartProps = Intergalactic.InternalTypings.EfficientOmit<
   tooltipViewType?: 'all' | 'single';
   /** Show percent value in tooltip */
   showPercentValueInTooltip?: boolean;
+  /** Custom percent formatter. */
+  percentFormatter?: (value: number) => number;
```
ilyabrower, line 22: so, the type should be in Abstract too

semcore/d3-chart/__tests__/index.test.tsx
slizhevskyv-semrush: yeap, good point!

semcore/d3-chart/src/component/Chart/CigaretteChart.tsx
slizhevskyv-semrush: looks like it's better to move it to `Cigarette` and as soon as we decide to have this possibility for other charts, move it to abstract layer...
what do you think?

ilyabrower (dismissed):

j-mnizhek:
@Valeria-Zimnitskaya Lera, I have a proposal to name the values for this new prop:

`round` | `floor` | `ceil` | `oneDecimal` | `twoDecimals`

Or even separate it into two props like:
`roundingMode`: `round` | `floor` | `ceil`
`precision`: `0` | `oneDecimal` | `twoDecimals`


What do you think? Looks like `toFixed1` and `toFixed2` could be tough to understand for people with less technical background.

j-mnizhek (approved):

Valeria-Zimnitskaya (approved):

PR #2830, opened 2026-04-01
UIK-5034/rewrite component to ts

slizhevskyv-semrush:
#### Fixed

- Rewrite component to TS.

## Motivation and Context

Nothing changed in how component works, just rewrite to TS.

## How has this been tested?

Manually

## Types of changes

- [x] Nice improve.

## Checklist:


semcore/radio/src/index.js
slizhevskyv-semrush, line 1: @ilyabrower just noticed that the old implementation doesn't export `inputProps`.

What do you think, shall we add it?

semcore/accordion/src/Accordion.tsx
```diff
@@ -172,23 +171,40 @@ class Toggle extends Component<
...
+    typeof RootItem,
+    'ToggleButton'
+  >,
+) {
   const { styles } = props;
 
-  const SToggleButton = Root;
   return sstyled(styles)(
-    <SToggleButton alignItems='center' render={Flex} role='button' {...props} />,
+    <Root render={Flex} {...props} alignItems='center' role='button' />,
```
ilyabrower: not sure about this changes in order of props...

semcore/errors/src/PageError/PageError.tsx
```diff
@@ -28,8 +31,10 @@ class RootPageError extends Component<PageErrorProps, typeof RootPageError.enhan
   render() {
     const { Children, getI18nText, onClick, titleTag, ...other } = this.asProps;
 
+    type A = typeof other['ref'];
```
ilyabrower: ?

semcore/errors/src/PageError/PageError.tsx
```diff
@@ -28,8 +31,10 @@ class RootPageError extends Component<PageErrorProps, typeof RootPageError.enhan
...
+
     return (
-      <Error {...other}>
+      <Root render={Error} {...other}>
```
ilyabrower: not sure about Root here... we didn't pass `Children, getI18nText, onClick, titleTag` this props to it before

semcore/radio/src/index.js
ilyabrower: inputProps? let's talk about it

semcore/accordion/src/Accordion.tsx
slizhevskyv-semrush: Can you imagine any realistic scenarios where these two attributes would need to change?

Looks like it should always have `role=button` and remain center-aligned :) 

Also, I’m not a big fan of prop spreading `{...props}` in this case. If you agree, we could consider removing that behavior altogether.

More broadly, what’s the actual value of `{...props}` here? For example, if something like `tabIndex={0}` is passed from the consumer, wouldn’t it still be applied to the underlying element even without explicit prop spreading?

semcore/radio/src/index.js
slizhevskyv-semrush: That's interesting, since the TS interface provides only `wrapRadio`, `Radio`, `RadioGroup` imports.
But, the `index.js` exports everything from the source file _including_ `inputProps`. 
It means if your project is on ts, you won't see that `inputProps` exists, and vice versa for `js` :D

semcore/accordion/src/Accordion.tsx
ilyabrower: let's just remove {...props}

ilyabrower (approved):

PR #2837, opened 2026-04-02
[UIK-5053][data-table] fixed stories with checkboxes

ilyabrower:
## Types of changes

- [x] Nice improve.

## Checklist:


slizhevskyv-semrush (approved):

stories/components/data-table/advanced/examples/checkbox_in_big_table/index.tsx
```diff
@@ -52,9 +42,6 @@ const Demo = (props: CheckboxExampleProps) => {
           scrollPaddingTop: '44px',
         }}
       >
-        <ScreenReaderOnly role='status' aria-live='polite'>
```
sheila-semrush, line 55: this shouldn't be removed from this example, the text just need to be slightly different: "Actions are available before the table" (btw this exact pattern is used in the example we're currently testing, and the participant said the announcement was very convenient)

stories/components/data-table/advanced/examples/checkbox_in_big_table_reactive/index.tsx
```diff
@@ -19,46 +19,14 @@ type CheckboxExampleProps = {
...
   const handleDeselectAll = () => {
     selectedRows.clearAll();
     tableRef.current?.focus();
   };
 
-  React.useEffect(() => {
-    const unsubscribe = selectedRows.subscribe(SelectableRows.TOGGLE_EVENT, () => {
-      const selectedRowsSize = selectedRows.get().length;
-      if (selectedRowsSize > 0) {
-        setAriaMessage('Action bar appeared before the table');
```
sheila-semrush, line 35: same - we need to keep the behavior, just change the text

stories/components/data-table/docs/examples/checkbox-in-table.tsx
```diff
@@ -1,35 +1,24 @@
...
-  const [selectedRowsDisplay, setSelectedRowsDisplay] = React.useState(0);
-  const [ariaMessage, setAriaMessage] = React.useState('');
+  const { count } = useSelectedRowsCount(selectedRows);
   const [currentPage, setCurrentPage] = React.useState(0);
   const tableRef = React.useRef<HTMLDivElement>(null);
 
-  const handleChangeSelectedRows = (value: string[]) => {
-    setSelectedRows(value);
-    if (!selectedRows.length) setAriaMessage('Action bar appeared before the table');
-    if (value.length) setSelectedRowsDisplay(value.length);
```
sheila-semrush, line 20: aria message now is announced every time selection changes (previously there was this logic to only set the aria message if the previous value was zero)

But, I think instead of getting back the old solution, we can do it slightly differently:
- change the text to "Actions are available before the table" (like in the other examples)
- remove the logic clearing the message after timeout. Clear the message only if selected rows length = 0. It's ok that user might stumble upon the message when navigating through the page - because now the text is neutral and appropriate even for that case.

stories/components/data-table/advanced/examples/checkbox_in_big_table/index.tsx
ilyabrower: I'm not sure, that we need it in advanced example, but if you insist, I'll put it back

stories/components/data-table/docs/examples/checkbox-in-table.tsx
ilyabrower: changed to your suggestion

stories/components/data-table/advanced/examples/checkbox_in_big_table/index.tsx
sheila-semrush: let's add it, this code can be copied/referenced as well, it's better to be accessible

ilyabrower: I removed this old example, and added SR for the reactive example (your next comment)

sheila-semrush (approved):

Valeria-Zimnitskaya (approved):

PR #2839, opened 2026-04-03
[UIK-5057][chore] add pipeline to release tools

slizhevskyv-semrush:
## Motivation and Context

Added GH pipeline to release tools.

## How has this been tested?

Manually

## Types of changes

- [ ] Nice improve.

## Checklist:


tools/continuous-delivery/index.ts
```diff
@@ -132,6 +132,40 @@ const sendReleaseChangelog = async (endpoints: string[]) => {
...
+
+    process.exit(1);
+  }
+
+  if (!availableTools.has(tool)) {
+    log(`Unable to find ${tool} within available tools...`);
+    log(`Available tools: \n${Array.from(availableTools.keys()).join('\n')}.`);
+
+    process.exit(1);
+  }
```
ilyabrower, line 143: what do you think, if we will use only this `if`. And will return a `Map` from `getAvailableTools` even if it'll have errors in json.parse ? Let's just add some log about error in json parsing there?

tools/continuous-delivery/index.ts
```diff
@@ -132,6 +132,40 @@ const sendReleaseChangelog = async (endpoints: string[]) => {
...
+
+  const toolRootPath = availableTools.get(tool)!;
```
ilyabrower: And also, I'd prefer to:
```
const toolRootPath = availableTools.get(tool);
if (!toolRootPath) {
 // logs from previous `if`
```
So, we don't have to use `!`

tools/continuous-delivery/index.ts
```diff
@@ -132,6 +132,40 @@ const sendReleaseChangelog = async (endpoints: string[]) => {
...
+  const packages = new Package();
+  await packages.collectToolPackageBy(toolRootPath);
+
+  const changelog = new Changelog(`tools/${tool}_`, prevReleaseVersion, packages.list);
+  await changelog.collectFromHistory();
+
+  await packages.updateVersions(changelog.data);
+  await gitUtils.initNewToolsRelease(changelog.data.version, packages.list);
+
+  NpmUtils.build([tool]);
```
ilyabrower: I think, this is a CI step, not a part of this job in pipeline...

tools/continuous-delivery/index.ts
slizhevskyv-semrush: sure, I'll update it :)

tools/continuous-delivery/index.ts
slizhevskyv-semrush: make sense, thanks!

tools/continuous-delivery/index.ts
slizhevskyv-semrush: good point, will separate it :)

ilyabrower (approved):

PR #3015, opened 2026-06-19
[UIK-3876][docs] update Wizard documentation

j-mnizhek:
## Motivation and Context

1. Removed some info from the styling section in the Wizard docs.

2. Removed the custom substep example completely from both the docs and stories. I couldn't find a single use case for it in Figma over the last year, and honestly don't remember anyone asking for it recently.

3. Added "optional" text to the example with custom step.

4. Kept Lera's substep example in the test stories, just in case.

## How has this been tested?
Manually.

## Types of changes

- [X] Nice improve.

## Checklist:

- [X] I have updated the documentation accordingly.


website/docs/components/wizard/wizard.md
```diff
@@ -36,78 +36,37 @@ Component consists of the following:
 
 ## Appearance
 
-### Stepper states
-
-Table: Stepper states
-
-| State    | Appearance example       | Styles                                                                                         |
-| -------- | ------------------------ | ---------------------------------------------------------------------------------------------- |
-| Normal   | ![](static/normal.png)   | `background-color: var(--control-primary-advertising)`, `border-radius: var(--rounded-medium)` |
```
sheila-semrush, line 45: can we keep the list of the stepper states with images? we can remove the Styles column, but I think the list of states is useful, otherwise it's a bit tricky to understand which states exist and how they look

website/docs/components/wizard/wizard.md
```diff
@@ -36,78 +36,37 @@ Component consists of the following:
...
-| Hover    | ![](static/hover.png)    | `background-color: var(--control-primary-advertising-hover)`, `cursor: pointer`                |
-| Active   | ![](static/active.png)   | `background-color: var(--control-primary-advertising-active)`                                  |
-| Disabled | ![](static/disabled.png) | Use `--disabled-opacity` token.                                                                |
-| Checked  | ![](static/checked.png)  | The number changes to a `Check` icon with M size.                                              |
-
-### Stepper styles
-
-If a `Stepper` has optional text, it should have the following styles:
```
sheila-semrush, line 53: "optional" btw is not part of the component, unfortunately...

website/docs/components/wizard/wizard.md
```diff
@@ -36,78 +36,37 @@ Component consists of the following:
...
-margin-top: var(--intergalactic-spacing-1x);
-```
-
-### Content area styles
-
-#### Header margin
-
-![margin-bottom: 20px; color: var(--gray-800); font-size: var(--fs-500); line-height: var(--lh-500); font-weight: 700;](static/header.png)
-
-#### Footer margins
```
sheila-semrush, line 66: footer is not in the component either( all sizes and margins should be custom set (only stepBack and stepNext have default size L, other controls like CTA buttons will have size set in place)

j-mnizhek: Blin! I was sure that it's already in the component 🥲 I'll bring it back

sheila-semrush (approved):

Valeria-Zimnitskaya (approved):

PR #3129, opened 2026-08-14
[UIK-5742][side-panel] redesign

ilyabrower:
## Types of changes

- [ ] Nice improve.

## Checklist:


slizhevskyv-semrush (approved):

semcore/side-panel/src/SidePanel.tsx
```diff
@@ -248,14 +248,10 @@ function Back(
...
-  const { Children, styles } = props;
+  const { styles } = props;
 
   return sstyled(styles)(
-    <SBack render={ButtonLink} color='text-hint' size={100} addonLeft={ArrowLeft}>
-      <ButtonLink.Text ellipsis>
-        <Children />
-      </ButtonLink.Text>
-    </SBack>,
+    <SBack render={ButtonLink} color='text-hint' size={100} addonLeft={ArrowLeft} />,
```
Valeria-Zimnitskaya: @ilyabrower @sheila-semrush i think,  it should have an accessible name, aria-label

Valeria-Zimnitskaya: @sheila-semrush i'm not sure icon color is correct here 
Figma: 
Image

Valeria-Zimnitskaya: @sheila-semrush In figma i see size =300, which is correct?

Image

semcore/side-panel/src/style/side-panel.shadow.css
```diff
@@ -61,40 +61,40 @@ SClose {
...
 SBack {
   max-width: 100%;
 }
 
 SHeader {
-  padding-left: var(--intergalactic-spacing-6x, 24px);
-  margin-bottom: var(--intergalactic-spacing-1x, 4px);
-  padding-right: calc(var(--intergalactic-spacing-10x, 40px) + var(--intergalactic-spacing-4x, 16px));
-  margin-top: var(--intergalactic-spacing-1x, 4px);
+  padding-left: var(--intergalactic-spacing-content-padding-xlarge-extended, 24px);
```
Valeria-Zimnitskaya, line 72: @sheila-semrush  gap is missed here?

Image

semcore/side-panel/src/style/side-panel.shadow.css
```diff
@@ -61,40 +61,40 @@ SClose {
...
+  padding-bottom: var(--intergalactic-spacing-content-padding-xsmall, 4px);
+  padding-right: var(--intergalactic-spacing-14x, 56px);
+  padding-top: var(--intergalactic-spacing-content-padding-xsmall, 4px);
   min-height: calc(var(--intergalactic-spacing-10x, 40px) + var(--intergalactic-spacing-3x, 12px));
-  box-shadow: inset 0 -1px 0 var(--intergalactic-border-primary, oklch(0.137 0.026 175.7 / 0.161));
+  border-bottom: 1px solid var(--intergalactic-border-primary, oklch(0.137 0.026 175.7 / 0.161));
   box-sizing: border-box;
   flex-shrink: 0;
+  flex-direction: column;
+  justify-content: center;
```
Valeria-Zimnitskaya, line 81: @sheila-semrush is that expected? because in figma i see align-items=center

Image

semcore/side-panel/src/SidePanel.tsx
```diff
@@ -248,14 +248,10 @@ function Back(
   props: Intergalactic.InternalTypings.InferComponentProps<NSSidePanel.Back.Component>,
 ) {
   const SBack = Root;
-  const { Children, styles } = props;
+  const { styles } = props;
 
   return sstyled(styles)(
-    <SBack render={ButtonLink} color='text-hint' size={100} addonLeft={ArrowLeft}>
```
sheila-semrush, line 254: @ilyabrower if we deprecate the Back button and replace with a buttonlink in the examples, maybe we should keep its old code?

semcore/side-panel/src/SidePanel.tsx
sheila-semrush: the color in figma was incorrect, I fixed it. But it shouldn't be color in code also, it should be use=secondary. I fixed it in the example, but we might not want to fix it here because we might not want to update the back button at all. Left another comment about it above.

sheila-semrush: nope, I fixed it in the example. About here - see other comments

semcore/side-panel/src/style/side-panel.shadow.css
sheila-semrush: we agreed to implement the Back button and all corresponding styles only in the example, including gap (it's implemented as margin there, which is ok)

semcore/side-panel/src/SidePanel.tsx
sheila-semrush: not necessarily, see comment above

semcore/side-panel/src/style/side-panel.shadow.css
sheila-semrush: the structure is different in Figma & implementation, so this is fine

semcore/side-panel/src/SidePanel.tsx
ilyabrower: I think no, we shouldn't... why do you think so?

sheila-semrush (approved):

Valeria-Zimnitskaya (approved):

PR #3145, opened 2026-08-20
[UIK-5776][data-table] improved cells rerenders

ilyabrower:
#### Fixed

- Unnecessary rerenders for dataTable cells.

## Types of changes

- [X] Nice improve.

## Checklist:


semcore/data-table/src/components/Head/Column.tsx
```diff
@@ -64,7 +64,13 @@ export class Column<
...
 
   componentDidMount() {
-    const { parent, sticky, changeSortSize, name, sort, scrollDirection } = this.asProps;
+    const { parent, sticky, changeSortSize, name, sort, scrollDirection, headerNodesMap } = this.asProps;
+    const columnElement = this.columnRef.current;
+    const columnName = columnElement?.getAttribute('name');
+
+    if (columnElement && columnName) {
+      headerNodesMap.set(columnName, columnElement);
+    }
```
slizhevskyv-semrush, line 73: let's free up the elements in `componentWillUnmount` (suppose in `DataTable.tsx`...)

semcore/data-table/src/components/Head/Head.tsx
```diff
@@ -56,39 +104,22 @@ class HeadRoot<
...
     return {
       use,
       gridArea: gridAreaGroupMap.get(index),
       withConfig: children === undefined,
       getFixedStyle,
       shadowVertical: (firstColumn.showShadowVertical || lastColumn.showShadowVertical) ? shadowVertical : undefined,
-      style,
+      style: this.columnStyle.get(props.name),
       scrollDirection,
+      // top: top && scrollDirection !== 'horizontal' ? `${top}px` : undefined,
```
slizhevskyv-semrush: do you need this line? :)

slizhevskyv-semrush (approved):

semcore/data-table/src/components/Body/Row.tsx
Valeria-Zimnitskaya: @ilyabrower  Am I correct in understanding that this lines Row.tsx:584-592 are left over and no longer used?

recalculateCellStyle above (lines 98–115) already does the same thing and stores the result in this.cellStyle, while line 609 uses style={this.cellStyle.get(column.name)}. So it looks like this local style is written to but never read.

Could you please check?

semcore/data-table/src/components/Head/Column.tsx
ilyabrower: I added it here, because the Column can be changed, but the table could be still the same...

semcore/data-table/src/components/Head/Head.tsx
ilyabrower: so as not to forget) removed comment

semcore/data-table/src/components/Body/Row.tsx
ilyabrower: yes, this is redundant, thank you

Valeria-Zimnitskaya (approved):

PR #3147, opened 2026-08-20
[UIK-5777][core] prevent calling redundat layout/rerender calls for useScrollBarWidth hook

slizhevskyv-semrush:
#### Fixed

- Performance regression for `useScrollBarWidth` hook.

## Motivation and Context

The hook now uses a single module-level `resize` listener and cached measurement shared across all instances (via `useSyncExternalStore`) instead of running its own `resize` listener, forced layout read and `setState` on every mount. Eliminates N duplicate `calculateScrollBar` calls where N is the number of components using the hook (e.g. `Tooltip`). No API changes.

## How has this been tested?

Manually

## Types of changes

- [x] Nice improve.

## Checklist:


semcore/core/src/utils/use/useScrollBarWidth.ts
```diff
@@ -1,39 +1,66 @@
...
-  const af = useRef<number | null>(null);
+type Dimension = 'vertical' | 'horizontal';
+
+const state: Record<Dimension, number> = { vertical: 0, horizontal: 0 };
+const listeners = new Set<() => void>();
+
+let rafId: number | null = null;
+let inited = false;
+
+function measureAndNotifyIfNeeded() {
```
ilyabrower: maybe just measure? =)

ilyabrower (approved):

semcore/core/src/utils/use/useScrollBarWidth.ts
slizhevskyv-semrush: I like such long names :D  but, yeah. it could be just measure :)

semcore/core/src/utils/use/useScrollBarWidth.ts
```diff
@@ -1,39 +1,66 @@
-import { useEffect, useState, useRef } from 'react';
+import { useSyncExternalStore } from 'react';
```
Valeria-Zimnitskaya: @slizhevskyv-semrush useSyncExternalStore was introduced in React 18, while @semcore/core currently declares support for React 16.8–19. Importing it directly from react will break consumers using React 16 or 17.
Could we use the backwards-compatible something? (if i correct))

slizhevskyv-semrush: yeah, good point! Thanks!
My assumption was that we support the current version -1. I will take care about it :)

Valeria-Zimnitskaya (approved):

